PR #21, opened 2017-01-09
Rewrite run.cmd and write some bash scripts for it.

romandev:
From now, we can use bash script on all platforms from now because we already
applied win-bash to our project(#17). So, rewrite run.cmd to support bash script
on all platforms and then add/write some scripts to work as before.


hyungheo:
This idea is pretty good

PR #25, opened 2017-01-11
Add downloading nodejs binary for platform

JayChl:
script for downloding nodejs binary for linux platform. 
This patch does not support downloading windows because currently we can not unzip file on win-bash

### DIFF
--- a/scripts/get_nodejs_bin
+++ b/scripts/get_nodejs_bin
@@ -80,26 +80,24 @@ function donwload_file_to_path() {
 # $2 : path of uncompressed files
 # Remarks : if $2 is not given, uncompress files to current path
 function extract_archive() {
-    if [ -z "$2" ]; then
-      dest="./"
-    else
-      dest="$2"
-    fi
+  if [ -z "$2" ]; then
+    dest="./"
+  else
+    dest="$2"
+  fi
 
-    if [ -f "$1" ] ; then
-      # NAME=${1%.*}
-      # mkdir $NAME && cd $NAME
-      case $1 in
-        *.tar.gz) tar xvzf $1 -C $dest ;;
-        *.gz) gunzip $1 > $dest ;;
-        *.tar) tar xvf $1 -C $dest ;;
-        # FIXME(Jaechul): Current our win-bash doens't support unzip!
-        # So this woudn't run on windows
-        *.zip) unzip $1 -d $dest ;;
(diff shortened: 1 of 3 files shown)


scripts/get_nodejs_bin
```diff
@@ -0,0 +1,105 @@
...
+}
+
+# extract archive files to given path
+#
+# Arguments:
+# $1 : path of archive
+# $2 : path of uncompressed files
+# Remarks : if $2 is not given, uncompress files to current path
+function extract_archive() {
+    if [ -z "$2" ]; then
```
romandev: nit: Fix indentation (4->2)

You are already using 2-spaces indentation in other functions.

romandev: These code(83line ~ 87line) can be simplified as follows:
  - dest="${2-./}"

But it's not mandatory. If you think your code is better for readability, you don't need to fix this.

scripts/run
```diff
@@ -2,6 +2,7 @@
 
 source ./scripts/path_info
 source ./scripts/platform_info
+source ./scripts/get_nodejs_bin
```
romandev: If possible, I think it's better to keep alphabetical order. :)

scripts/get_nodejs_bin
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
+NODEJS_7x_LATEST_URL="https://nodejs.org/dist/latest-v7.x"
+NODEJS_WIN_x64="win-x64"
+NODEJS_WIN_x86="win-x86"
+NODEJS_LINUX_x64="linux-x64"
+NODEJS_LINUX_x86="linux-x86"
+# TODO : This is temporal fixed version. Need to get latest or specific version from dist repository
+NODEJS_LATEST7X_VERSION="node-v7.4.0"
+
+source ./scripts/path_info
```
romandev: I think it's better to move this line to top of this file.

scripts/get_nodejs_bin
```diff
@@ -0,0 +1,105 @@
...
+# Remarks : if $2 is not given, uncompress files to current path
+function extract_archive() {
+    if [ -z "$2" ]; then
+      dest="./"
+    else
+      dest="$2"
+    fi
+
+    if [ -f "$1" ] ; then
+      # NAME=${1%.*}
```
romandev: Do we need these comments?

PR #34, opened 2017-01-14
Firebase connect and disconnet

daehyunjimmy:
User use below call
* **fcm.connect()**
 1. Connect Firebase and get token
 2. Add event Handler to push notification

* **fcm.disconnect()**
 1. Disconnect Firebase and delete token


romandev:
@daehyunjimmy Your patch looks good to me but please split this into smaller patches for easier review.

PR #37, opened 2017-01-15
Fix a bug that the run.cmd is failed on Mac OS.

romandev:
Current implementation doesn't support for Mac OS yet. So, this patch is
implementing it and then adding some utility functions for it.
Also, this patch is fixing some of typos in existing codes.


romandev:
Hi @JayChl 

Could you please review this patch?

hyungheo:
It's working well

romandev:
Hi guys,

Thank you for your review!

PR #42, opened 2017-01-18
Refactor the server related codes.

romandev:
This patch includes the following things.
 - Introduce config.js for server configuration.
 - Factor out https_server related codes to separated module. (https_server)
 - Move the certification key files to out of server directory.
 - Remove unnecessary codes. (routes/*, app.js)
 - Set |client| directory for static directory.


romandev:
PTAL

PR #53, opened 2017-01-24
Linking to travis system for absolute project

gogag1:
I linked to travis system for absolute project.
You can check result at https://travis-ci.org/.
And you can refer mocha test at https://mochajs.org.
Now testing will work on server and client folders.

Issue: Test automation #8


.travis.yml
```diff
@@ -0,0 +1,9 @@
+language: node_js
+node_js: 
+  - "node"
+install:
+  - npm install -g mocha
```
romandev: Do we really need to use -g option? We might not want global installation.

CoconutPerm: I think that is not a problem because of it is travis's option. isn't it?

gogag1: yes right, we don't need global installation.
But the installation doesn't affect our project. 
It will be installed on travis system while running test.

gogag1: CoconutPerm@
yes right. it is for travis.

PR #60, opened 2017-02-03
Download node platform and unzip

coconutperm:
We need download node platform and unzip specific path.

### DIFF
--- a/scripts/platform_info
+++ b/scripts/platform_info
@@ -21,7 +21,7 @@ get_platform_name() {
     # FIXME(zino): Once we finish to implement NodeJS Downloader via script,
     # the following platform name should be |windows_x86_64|.
     # Please see: https://github.com/romandev/absolute/issues/19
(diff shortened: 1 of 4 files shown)


scripts/platform_info
```diff
@@ -21,7 +21,7 @@ get_platform_name() {
     # FIXME(zino): Once we finish to implement NodeJS Downloader via script,
     # the following platform name should be |windows_x86_64|.
     # Please see: https://github.com/romandev/absolute/issues/19
```
JayChl: Let's remove this FIXME, so this issue is closed

scripts/run
```diff
@@ -10,8 +10,8 @@ function run_server() {
   # FIXME(zino): Once we finish to implement NodeJS Downloader via script,
```
JayChl: Lets remove FIXME comment

JayChl:
Looks good to me

PR #62, opened 2017-02-03
Remove included platform binary

coconutperm:
We need remove included platform binary due to script is changed to
download binary to server.


romandev:
lgtm. This is a great work!
Thanks.

PR #65, opened 2017-02-03
fix the failure of download node and db when running run.cmd

gitdongguri:


scripts/util
```diff
@@ -50,7 +50,7 @@ function download_file_to_path() {
 
   if [ ! -d "$path" ]; then
     echo "targetPath does not exist. create $path"
-    mkdir $path
```
hyungheo: looks good to me

PR #68, opened 2017-02-05
added eslint and gulp modules in package.json

gitdongguri:
This patch is for supporting eslint feature using gulp and travis.
Before using eslint and gulp modules, we have to define the dependency in package.json.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,28 @@
+module.exports = {
+    "env": {
+        "browser": true,
+        "es6": true
+    },
+    "extends": "airbnb-base",
+    "parserOptions": {
+        "ecmaFeatures": {
+            "jsx": true
+        },
+        "sourceType": "module"
+    },
+    "rules": {
+        "indent": [
+            "error",
+            4
+        ],
+        "quotes": [
+            "error",
+            "single"
+        ],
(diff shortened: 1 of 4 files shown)


.eslintrc.js
```diff
@@ -0,0 +1,28 @@
+module.exports = {
+    "env": {
```
coconutperm: Tab is 2 space. right?

scripts/run_gulp
```diff
@@ -0,0 +1,20 @@
...
+source ./scripts/util
+
+function run_gulp() {
+  platform_name=$(get_platform_name)
+  
+  if $(is_windows); then
+    #TODO(gitdongguri): run gulp in windows
+    echo "ttt"
+  else
+  echo "eeee"
```
coconutperm: plz check indent

.eslintrc.js
gitdongguri: Okay! How can I modify patch that I've done pull request?..

romandev: You can push additional changes into the same branch. :)

scripts/run_gulp
gitdongguri: indent is right but useless log is the problem right?

PR #73, opened 2017-02-16
Refactor scripts for bootstrap.

romandev:
WIP

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
-node_modules
-platform
 *.swp
-*.tar.gz
-*.zip
-npm-debug.log
+node_modules
+third_party/mongodb
+third_party/node
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,3 @@
 language: node_js
-node_js: 
-  - "node"
-install:
-  - npm install -g mocha
-env:
-  - TEST_DIR=server
-  - TEST_DIR=client
-script: cd $TEST_DIR && mocha ./test/*.js
+script:
+  - ./absolute platform_test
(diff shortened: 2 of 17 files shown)


bootstrap/sync_third_party.sh
```diff
@@ -0,0 +1,107 @@
...
+
+. bootstrap/common/platform_info.sh
+. bootstrap/common/util.sh
+
+function sync_node() {
+  local target_path="./third_party/node"
+  local base_url="https://nodejs.org/dist/v6.9.5"
+
+  case $(get_platform_name) in
+    windows_x86) local target_url="/node-v6.9.5-win-x86.zip" ;;
```
coconutperm: I think better that node version is defined to constant.

bootstrap/sync_third_party.sh
```diff
@@ -0,0 +1,107 @@
...
+
+  sync_third_party $base_url$target_url $target_path
+}
+
+function sync_mongodb() {
+  local target_path="./third_party/mongodb"
+  local base_url="http://downloads.mongodb.org"
+
+  case $(get_platform_name) in
+    windows_x86) local target_url="/win32/mongodb-win32-i686-3.2.12.zip" ;;
```
coconutperm: It also above comment.

bootstrap/sync_third_party.sh
romandev: I think it's unnecessary. because we can not sure the link is always available.
We should replace the hard coded values with configuration files or deps files.

bootstrap/sync_third_party.sh
romandev: ditto

PR #75, opened 2017-02-21
Apply new solution instead of forever

coconutperm:
I think pm2 better than forever and nodemon.
It support daemonize and auto-reload.

### DIFF
--- a/server/pm2_config.json
+++ b/server/pm2_config.json
@@ -1,0 +1,17 @@
+{
+  "apps" : [{
+    "name"        : "absolute",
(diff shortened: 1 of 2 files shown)


server/pm2_config.json
```diff
@@ -0,0 +1,22 @@
+{
```
romandev: Do we really have to have this configuration file?
Too much configuration files might be bad sign to us.

coconutperm: most options are for log property and watch options.

PR #98, opened 2017-02-28
Apply mongodb script

coconutperm:
We need run mongodb as child process


server/database.js
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2017 The Absolute Authors. All rights reserved.
```
romandev: Please move this file to bootstrap/command/start_db

And you can use shebang as follows:
#/usr/bin/env node

And you can set 755 permission.

PR #118, opened 2017-03-01
Add implementation of register call

daehyunjimmy:
* implementation of register API
 - register Service Worker
 - subscribe push manager 
 - send push description to server


client/push/sw_manager.js
```diff
@@ -17,6 +40,18 @@
     }
 
     function sendPushTokenToServer() {
+      var http = new XMLHttpRequest();
+      var url = "https://localhost:9443/api/push/client";
```
JayChl: Lets add "TODO : " to make sure we will change this url from config or running server url

client/push/sw_manager.js
```diff
@@ -17,6 +40,18 @@
     }
 
     function sendPushTokenToServer() {
+      var http = new XMLHttpRequest();
+      var url = "https://nadonguri.com/api/push/client";
```
romandev: Shouldn't be '/api/push/subscribe'?

client/push/sw_manager.js
```diff
@@ -8,6 +8,29 @@
     var swRegistration = '';
```
romandev: Why do we need this?

PR #119, opened 2017-03-01
Add API to control user mobile camera device.

cs-lee:
Add API to control user mobile camera device.
- static resetOrientation( ) : Set correct image orientation.
- static getOrientation( ) : Get orientation value from captured image.
- public setImage( ) : Set image from camera.

### DIFF
--- a/client/device/camera/camera.js
+++ b/client/device/camera/camera.js
@@ -1,0 +1,23 @@
+(function (window) {
+  /**
+   * Set correct image orientation.
+   */
+  var resetOrientation = function (imgSrc, orientation, callback) {
+  };
+
+  /**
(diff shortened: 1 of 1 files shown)


client/device/camera/camera.js
```diff
@@ -0,0 +1,23 @@
...
+   * Get orientation value from captured image.
+   */
+  var getOrientation = function (file, callback) {
+  };
+
+  /**
+   * API for control user mobile camera device.
+   */
+  var camera = {
+    setImage : function (inputId, callback) {
```
romandev: As we discussed, this function's signature should be func(parentElement, callback).

But please do it in the follow-up CL.

PR #123, opened 2017-03-01
Add rest api for order menu.

coconutperm:
We need rest api for order menu.
https://{address}/api/order (POST)


server/order/router/order_router.js
```diff
@@ -0,0 +1,11 @@
+const express = require('express');
```
JayChl: Need require express?

PR #125, opened 2017-03-01
Implement deep link to open an external browser.

gogag1:
Implement deep link to open an external browser.
KakaoTalk is using internal WebView to open link.
But we can't hand service worker and related features.
So, when we open link in kakao talk, we need to open link using chrome or sbrowser.

BUG=https://github.com/romandev/absolute/issues/96

### DIFF
--- a/client/deeplink/deeplink.js
+++ b/client/deeplink/deeplink.js
@@ -1,15 +1,16 @@
-(function () {
+(function (window) {
   const deeplink = {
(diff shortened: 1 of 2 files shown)


client/deeplink.js
```diff
@@ -0,0 +1,15 @@
+(function () {
```
romandev: This file should be located in client/deeplink/deeplink.js

Issue #77, opened 2017-02-23
Server does not work when start in Window Git Bash.


cs-lee:
Server does not work when start in Window Git Bash.
1. The resone of that is pwd.exe can not support hangul.

romandev:
I didn't know that. Nice catch! Please follow it up.

cs-lee:
This issue resolved.

PR #141, opened 2017-03-02
Bug fix - type command does not work in window. (#77)

cs-lee:
1. The "type" command does not work when call it from "third_party/win-bash/bash" in window OS. Because "third_party/win-bash" path not in PATH env. So "type" can't find argument command always except builtin command.
2. "if" condition fix in download function to use wget when only can use wget command.

After this path "absolute start" command work in window git bash.

### DIFF
--- a/bootstrap/common/util.sh
+++ b/bootstrap/common/util.sh
@@ -28,7 +28,7 @@ function download() {
   fi
 
   echo "Downloading... $url"
-  if is_windows_platform || can_use_command wget; then
+  if is_windows_platform && can_use_command wget; then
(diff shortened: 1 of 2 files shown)


bootstrap/common/util.sh
```diff
@@ -28,7 +28,7 @@ function download() {
   fi
 
   echo "Downloading... $url"
-  if is_windows_platform || can_use_command wget; then
+  if is_windows_platform && can_use_command wget; then
```
romandev: I think, this is not correct.
We want to use wget in Linux.

cs-lee: I will fix it and upload it again.

PR #143, opened 2017-03-03
Call register api to install service worker and subscribe push

daehyunjimmy:
1.  call register when index.html called
2. fix service worker not worked issue

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -9,6 +9,7 @@
 <body>
   <script>
     deeplink.createBrowser('com.android.chrome');
+    register();
   </script>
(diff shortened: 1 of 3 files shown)


client/index.html
```diff
@@ -9,6 +9,7 @@
 <body>
   <script>
     deeplink.createBrowser('com.android.chrome');
+    register();
```
romandev: I think this is not correct.
This function should be called in client/wedding/index.html
/cc @nadongguri

romandev:
Also , please fix lint error.

PR #144, opened 2017-03-04
Implement uuid generator rest api

coconutperm:
We need unique key for distributed system.
So implement rest api for generate uuid.


server/user/router/uuid_router.js
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2017 The Absolute Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+const express = require('express');
+const uuidGenerator = require('../controller/uuid_generator');
+
+const router = express.Router();
+
+router.post('/uuid', (req, res) => {
```
romandev: This should be 'get' method.

romandev: Please use request, response instead of req, res

package.json
```diff
@@ -9,7 +9,8 @@
     "forever": "~0.15.3",
     "body-parser": "~1.16.1",
     "mongoose": "4.8.5",
-    "request": "~2.16.2"
+    "request": "~2.16.2",
+    "node-uuid": "1.4.7"
```
romandev: You should use uuid instead of node-uuid.
It's a deprecated as I know.

PR #145, opened 2017-03-04
Implement push endpoint feature

coconutperm:
Define push endpoint schema and implement save endpoint rest api.
It should save endpoint with userId.


server/push/router/push_router.js
```diff
@@ -1,15 +1,20 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
 const express = require('express');
-const pushManager = require('./push_manager.js');
+const pushController = require('../controller/push_controller');
 
 const router = express.Router();
+
 router.post('/client', (req, res) => {
```
romandev: Please don't use req, res. We have to use full name instead of these.
Like request, response.

server/order/router/product_router.js
```diff
@@ -9,7 +9,7 @@ const router = express.Router();
 
 router.post('/add', (req, res) => {
```
romandev: Please don't use req, res. We have to use full name instead of these.
Like request, response.

server/order/router/product_router.js
```diff
@@ -9,7 +9,7 @@ const router = express.Router();
...
   productController.addProduct(JSON.stringify(req.body));
-  res.send('OK');
+  res.sendStatus(200);
 });
 
 router.post('/list', (req, res) => {
```
romandev: Please don't use req, res. We have to use full name instead of these.
Like request, response.

server/order/router/order_router.js
```diff
@@ -8,8 +8,12 @@ const orderController = require('../controller/order_controller');
 const router = express.Router();
 
 router.post('/', (req, res) => {
```
romandev: Please don't use req, res. We have to use full name instead of these.
Like request, response.

server/order/router/product_router.js
romandev: This should be 'get'.

server/order/router/order_router.js
romandev: Could you please explain what it is for?

server/push/router/push_router.js
```diff
@@ -24,7 +29,16 @@ router.delete('/client', (req, res) => {
...
     // Remove token from DB
-    pushManager.removeClient(clientToken);
+    pushController.removeClient(clientToken);
+    res.sendStatus(200);
+  } else {
+    res.sendStatus(400);
+  }
+});
+
+router.post('/add', (req, res) => {
```
romandev: What is a difference between  /client and /add?

server/server.js
```diff
@@ -21,7 +21,7 @@ app.use(express.static(path.join(__dirname, '../client')));
 
 app.use('/api/order', require('./order/router/order_router'));
 app.use('/api/product', require('./order/router/product_router'));
-app.use('/api/push/', require('./push/router.js'));
+app.use('/api/push/', require('./push/router/push_router'));
```
JayChl: It's under /push/router/ ,  the new name of "push_router.js" seems repeated

PR #146, opened 2017-03-04
Change database name

coconutperm:
Change name to absolute from test(default)

### DIFF
--- a/server/db/db_connect.js
+++ b/server/db/db_connect.js
@@ -13,7 +13,7 @@ function connectServer() {
     console.log('Mongoose disconnected');
   });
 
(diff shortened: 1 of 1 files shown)


server/db/db_connect.js
```diff
@@ -13,7 +13,7 @@ function connectServer() {
     console.log('Mongoose disconnected');
   });
 
-  mongoose.connect('localhost:27017', (err) => {
+  mongoose.connect('localhost:27017/absolute', (err) => {
```
romandev: Please do not use 'err'. Instead please use full name like 'error'.

PR #149, opened 2017-03-05
Refacor api

coconutperm:
Check product name and amount.

### DIFF
--- a/server/order/controller/product_controller.js
+++ b/server/order/controller/product_controller.js
@@ -5,8 +5,17 @@
 const Product = require('../model/product');
 
 exports.addProduct = function (productData) {
-  const product = new Product(JSON.parse(productData));
-  product.save();
+  return new Promise((resolve, reject) => {
+    Product.findOneAndUpdate({ name: productData.name }
+      , { amount: productData.amount, options: productData.options }
+      , { upsert: true }, (error, product) => {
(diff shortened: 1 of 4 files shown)


server/order/controller/product_controller.js
```diff
@@ -5,8 +5,17 @@
 const Product = require('../model/product');
 
 exports.addProduct = function (productData) {
-  const product = new Product(JSON.parse(productData));
-  product.save();
+  return new Promise((resolve, reject) => {
+    Product.findOneAndUpdate({ name: productData.name }
```
romandev: Is the name a unique key? If not so, this logic is incorrect.
Also, this is an add function. Then, do we really need to find something?

coconutperm: I think there are no product that has same name. aren't you?

coconutperm: I think there are no product that has same name. aren't you?
And It should use findOneAndUpdate with upsert option because of can't has same name product.

romandev: IMHO, the name can not be unique key. For example, there are two types of Americano, which has a different coffee bean. (This example is a real world example in coffee store in front of my house.)

But as-is okay for now. (in this branch) I hope we should fix it before cherry-picking master.

PR #150, opened 2017-03-05
Bug fix to run absolute in window git bash and cmd both.

cs-lee:
1. Change the "if" statement in the "download" function to determine whether
the command available.
2. Add binary path in PATH environment value for window.
3. Start script modify to determine which script used for launch in
absolute.sh.


hyungheo:
looks good to me

bootstrap/absolute.sh
```diff
@@ -10,18 +10,25 @@
 # Copy githooks to .git/hooks.
 cp $(githooks_path)/* $(absolute_path)/.git/hooks
 
-# Sync third_parties.
-sync_node
-sync_mongodb
-
 # Set path
+if is_windows_platform; then
+  if [ "$1" = "from-cmd" ]; then
```
romandev: I don't think we need this check.
Do we really need it?

romandev:
I left a inline comment. Please remove signed-off-by in commit description (We don't use it until now. So, you don't use git with -s option)
Also, I have two questions.
1. Does 'win-wget' is working well in both of 64bit system and 32bit system?
2. Shouldn't we have to put some license information to include the binary file in our project?

bootstrap/absolute.sh
cs-lee: That "if" statement is to export "$(third_party_path)/win-bash" path only user launch absolute in windows cmd. Without this line, bash-shell that already installed in user windows pc, run command in $(third_party_path)/win-bash.

PR #155, opened 2017-03-05
API:   getUuid()

hyungheo:
Syntax:
    getUuid()

Return:
    Promise Object

Description;
    Get uuid by using restful api (api/authorization/uuid) and save it on localStorage


client/authorization/authorization.js
```diff
@@ -0,0 +1,67 @@
...
+
+    // PUBLIC API
+    self.createUuid = function () {
+      return new Promise(function (resolve, reject) {
+        requestUuid().then(function (result) { return saveUuid(result); })
+                 .then(function (e) { resolve(e); });
+      });
+    };
+
+    self.getUuid = function () {
```
romandev: I think public API only needs getUUID.
The function will work as follows:
  If there is no uuid in local storage. Then create and return.
  If there is already uuid in local storage. Then read and return.

PR #160, opened 2017-03-06
API: order()

hyungheo:
Syntax:
    order(OrderInfo);
Pramemeter:
    OrderInfo {
      userId: String,
      items: OrderItem or Array[OrderItem,...],
      blob: Blob,
      option: String,
    }
    OrderItem {
      name: String,
      count: Number,
      amount: Number,
    }
Return:
    Promise
        resolve(e), reject(e)
        e: error_code (0: SUCCESS, 1: ERROR)
    }

Description:
    Send order to server and recieve error_code

### DIFF
--- a/client/order/order.js
+++ b/client/order/order.js
@@ -1,27 +1,132 @@
 (function (global, factory) {
-  /**
-   * Set correct image orientation.
-   */
   /*eslint-disable */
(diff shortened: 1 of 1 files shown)


romandev:
If you already checked it works well, then go ahead to merge.

Issue #96, opened 2017-02-28
[Hackathon] Implement deep link to open an external browser.


romandev:
When sending the message via KakaoTalk, the link will be opened on internal WebView.
Then, we can not control service worker and related features.
So, we have to implement and use deep link to open the url to an external browser.

gogag1:
![deep link](https://cloud.githubusercontent.com/assets/3260994/23413143/2bfc96da-fe1b-11e6-884f-40fce441e2f1.png)

PR #167, opened 2017-03-06
Deeplink should be sent when users are using andorid phone.

gogag1:
Deeplink should be sent when users are using andorid phone.
We blocked sending deeplink when users are using the others phones.

BUG=#96

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -8,13 +8,12 @@
 </head>
 <body>
   <script>
-    if(navigator.userAgent.indexOf("Android")) {
-      if((navigator.userAgent.indexOf("Chrome")==-1) ||
-      (navigator.userAgent.indexOf("Samsungbrowser")==-1))
-        deeplink.createBrowser('com.android.chrome');
-     }
-     else
-       window.location.href = 'https://www.nadongguri.com/wedding'
+    if(/Android/.test(navigator.userAgent)) {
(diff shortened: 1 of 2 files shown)


gogag1:
PTAL

client/index.html
```diff
@@ -8,7 +8,10 @@
 </head>
 <body>
   <script>
-    deeplink.createBrowser('com.android.chrome');
+    if(navigator.userAgent.indexOf("Android")>-1)
```
romandev: nit: if (/Android/.test(navgiator.userAgent))

gogag1:
I fixed displaying default browser. PTAL

client/index.html
```diff
@@ -8,7 +8,13 @@
 </head>
 <body>
   <script>
-    deeplink.createBrowser('com.android.chrome');
+    if(navigator.userAgent.indexOf("Android")) {
```
romandev: nit: if (/Android/.test(navigator.userAgent))

client/index.html
```diff
@@ -8,7 +8,13 @@
...
+      if((navigator.userAgent.indexOf("Chrome")==-1) ||
```
romandev: nit: /Samsung/.test(navigator.userAgent) (chrome as well)

PR #169, opened 2017-03-09
Deeplink should be passed when user are using webview.

gogag1:
Deeplink should be passed when user are using webview and 
not using chrome and samsungbrowser.

BUG = #94

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -8,7 +8,16 @@
 </head>
 <body>
   <script>
-    deeplink.createBrowser('com.android.chrome');
+	var ua = navigator.userAgent;
+	if(/Android/.test(ua)) {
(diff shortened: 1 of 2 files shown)


client/index.html
```diff
@@ -8,7 +8,16 @@
 </head>
 <body>
   <script>
-    deeplink.createBrowser('com.android.chrome');
+	var ua = navigator.userAgent;
```
romandev: This logic should be in deeplink.js

client/index.html
```diff
@@ -8,7 +8,16 @@
...
+	if(/Android/.test(ua)) {
```
romandev: How about making some utility functions for this checking?
For example..
isAndroid(), isChromiumWebView, isChromeBrowser, isSamsungBrowser.

PR #170, opened 2017-03-09
Deeplink should be sent when users are using android phone.

gogag1:
Deeplink should be sent when users are using android phone.
We blocked sending deeplink when users are using the others phones.

BUG=#96

### DIFF
--- a/client/deeplink/deeplink.js
+++ b/client/deeplink/deeplink.js
@@ -4,7 +4,6 @@
       const intentURI = [
         'intent://www.nadongguri.com/wedding#Intent',
         'scheme=https',
-        `package=${appPackage}`,
         'S.browser_fallback_url=http://www.nadongguri.com/wedding',
         'end',
       ].join(';');
--- a/client/deeplink/deeplink.js
+++ b/client/deeplink/deeplink.js
@@ -1,14 +1,41 @@
 (function (window) {
-  const deeplink = {
-    createBrowser(appPackage) {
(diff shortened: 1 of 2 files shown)


client/deeplink/deeplink.js
```diff
@@ -1,15 +1,41 @@
...
+
+    window.location.href = intentURI;
+  }
+  function isLatestWebView() {
+    return /Version\/[.0-9]* Chrome\/[.0-9]*/.test(navigator.userAgent);
+  }
+  function isAndroidDevice() {
+    return /Android/.test(navigator.userAgent);
+  }
+  function isSamsungBrowser() {
```
romandev: Maybe. the function name will be like isChromiumBasedBrowser()

client/deeplink/deeplink.js
```diff
@@ -1,15 +1,41 @@
...
+    return /Chrome/.test(navigator.userAgent);
+  }
+  function goToWeddingSite() {
+    window.location.href = 'http://nadongguri.com/wedding';
+  }
 
-      window.location.href = intentURI;
+  const deeplink = {
+    createBrowser() {
+      if (isAndroidDevice()) {
```
romandev: if (isAndroidDevice() && !isChromiumWebView() && isChromiumBasedBrowser()) {
    jumpToDefaultBrowser();
} else {
    redirectWeddingPage();
}

PR #175, opened 2017-03-16
use static message if notification data is not exist

JayChl:
This change is for wedding. 
Before applying new implementation getting notification data by user id, use static message

### DIFF
--- a/client/sw.js
+++ b/client/sw.js
@@ -21,7 +21,7 @@ self.addEventListener('push', function (event) {
     body = notificationOptions;
   }
 
-  self.registration.showNotification(title, body);
+  event.waitUntil(self.registration.showNotification(title, body));
 });
 
 self.addEventListener('notificationclick', function (event) {
--- a/client/sw.js
+++ b/client/sw.js
@@ -1,17 +1,41 @@
 /* eslint-disable */
 
 self.addEventListener('push', function (event) {
-  var data = {};
+
+  var title = "";
+  var body = "";
+
   if (event.data) {
+    var data = {};
+    console.log('get push notification data');
     data = event.data.json();
+    title = data.notification.title;
+    body = data.notification.body;
+  } else {
+    console.log('use static push notification data');
+
+    title = "나동진♡김소영 결혼을 축하해주세요!"
+    const notificationOptions = {
+      body: '2017년 4월 8일 토요일 오후 5시',
+      icon: './images/main.jpg',
+      badge: './images/badge-72x72.png',
+      tag: 'wedding',
+      url: 'https://nadongguri.com/wedding/'
+    };
+
+    body = notificationOptions;
   }
 
   event.waitUntil(
     self.registration.showNotification(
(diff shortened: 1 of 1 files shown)


client/sw.js
```diff
@@ -1,17 +1,41 @@
 /* eslint-disable */
 
 self.addEventListener('push', function (event) {
-  var data = {};
+
```
romandev: Please remove unnecessary line.

client/sw.js
```diff
@@ -1,17 +1,41 @@
...
+  var title = "";
```
romandev: var title; is enough
or we can declare these variables when it is actually used.

client/sw.js
```diff
@@ -1,17 +1,41 @@
...
+      title, body)
   );
+
 });
 
 self.addEventListener('notificationclick', function (event) {
-});
+  event.notification.close();
 
+  const clickResponsePromise = Promise.resolve();
```
romandev: Why do we need this?
I'm not sure but it looks unnecessary.

client/sw.js
```diff
@@ -1,17 +1,41 @@
 /* eslint-disable */
 
 self.addEventListener('push', function (event) {
-  var data = {};
+
+  var title = "";
+  var body = "";
+
   if (event.data) {
+    var data = {};
```
romandev: Let's merge this line and line11.

client/sw.js
```diff
@@ -1,17 +1,36 @@
...
     self.registration.showNotification(
-      data.notification.title, data.notification)
+      title, body)
   );
+
 });
 
 self.addEventListener('notificationclick', function (event) {
+  event.notification.close();
+  clickResponsePromise = clients.openWindow("https://nadongguri.com/wedding/");
```
romandev: Why do we need a global variable here? (clickResponsePromise)
I didn't find where the global variable is used.

client/sw.js
```diff
@@ -1,17 +1,36 @@
...
+      tag: 'wedding',
+      url: 'https://nadongguri.com/wedding/'
+    };
+    body = notificationOptions;
   }
 
   event.waitUntil(
     self.registration.showNotification(
-      data.notification.title, data.notification)
+      title, body)
```
romandev: We can make these two lines as one line. (and please append semicolon to end of line)
self.registration.showNotification(title, body);

client/sw.js
```diff
@@ -1,17 +1,36 @@
...
   );
+
```
romandev: nit: Please remove unnecessary blank line.

client/sw.js
JayChl: event.waitUntil(
    self.registration.showNotification(title, body));
or
  self.registration.showNotification(title, body);  
 without event.waitUntil?

client/sw.js
JayChl: This is unnecessary variable will remove this thanks

PR #180, opened 2017-03-27
Implemention of order page api

cs-lee:
1. Add api for order page.

### DIFF
--- a/client/order/orderui.js
+++ b/client/order/orderui.js
@@ -1,8 +1,8 @@
 (function (window) {
-  var orderFrame = document.createElement("iframe");
-  var iImg = document.createElement("img");
-  var iSend = document.createElement("button");
-  var iText = document.createElement("input");
+  var orderFrame = document.createElement('iframe');
+  var iImg = document.createElement('img');
+  var iSend = document.createElement('button');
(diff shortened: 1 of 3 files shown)


client/order/orderui.js
```diff
@@ -0,0 +1,101 @@
...
+  var iImg = document.createElement('img');
+  var iSend = document.createElement('button');
+  var iText = document.createElement('input');
+
+  /**
+   * API for control order ui
+   */
+  var orderPage = {
+    /**
+     * Initialization of rrder page.
```
romandev: What's rrder?

client/order/orderui.js
```diff
@@ -0,0 +1,101 @@
...
+     * @param String Text to be displayed on order page.
+     */
+    pageLoad(welcomeText) {
```
romandev: Why do we need to make this as public?

client/order/orderui.js
cs-lee: I will fix and commit again.

client/order/orderui.js
cs-lee: To set user parameters at page loading

PR #191, opened 2017-04-03
Syntax: orderList(uuid)

hyungheo:
Description:
    return Array of imagePath(static)
Usage:
   orderList(uuid).then(function(list) {
     ...
   });


romandev:
Done?

PR #215, opened 2017-04-21
add notification image

Wuseok:
add notification image


romandev:
We don't use hackathon branch anymore.

PR #224, opened 2017-05-15
Bug fix in sync_third_party( )

cs-lee:
trap command update argument so last command excute only
so tmp directory does not be removed.

### DIFF
--- a/bootstrap/common/sync_third_party.sh
+++ b/bootstrap/common/sync_third_party.sh
@@ -55,7 +55,8 @@ function sync_third_party() {
 
   # Install a trap to remove temporary path if this script is finished.
   # The trap will work well when unknown breaking happens like ctrl+c.
-  trap "rm -rf $temp_path" EXIT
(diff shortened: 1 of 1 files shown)


bootstrap/common/sync_third_party.sh
```diff
@@ -55,7 +55,8 @@ function sync_third_party() {
 
   # Install a trap to remove temporary path if this script is finished.
   # The trap will work well when unknown breaking happens like ctrl+c.
-  trap "rm -rf $temp_path" EXIT
+  export TRAP_COMMAND="rm -rf $temp_path;$TRAP_COMMAND"
```
romandev: Why do we need to export this command?
Also, shouldn't be |export TRAP_COMMAND="rm -rf $temp_path"|?

cs-lee: This is for remove all C:/tmp_@@@ directory. 
sync_third_party( ) function call twice. one is for node, another is for mongo.
but only one tmp directory remove. because when called second trap command, trap remove first one and apply second command. 
So I fix it by using env value.

PR #226, opened 2017-05-17
Apply body parser to server

JayChl:
- With body-parser we can parse incomming request

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -4,13 +4,20 @@
 
 const express = require('express');
 const path = require('path');
+const bodyParser = require('body-parser');
 
 const config = require('./config');
 const httpsServer = require('./https_server/https_server');
 const redirectServer = require('./https_server/redirect_server');
 
 const app = express();
+
 app.use(express.static(path.join(__dirname, '../client')));
+app.use(bodyParser.json({ limit: '10mb' }));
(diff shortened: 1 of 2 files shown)


nadongguri:
LGTM

server/server.js
```diff
@@ -4,13 +4,20 @@
 
 const express = require('express');
 const path = require('path');
+const bodyParser = require('body-parser');
```
romandev: I think we should use |import| instead of |require|.

server/server.js
```diff
@@ -4,13 +4,20 @@
...
 
 const config = require('./config');
 const httpsServer = require('./https_server/https_server');
 const redirectServer = require('./https_server/redirect_server');
 
 const app = express();
+
 app.use(express.static(path.join(__dirname, '../client')));
+app.use(bodyParser.json({ limit: '10mb' }));
```
romandev: I'm not sure if this limit is really needed in this branch.
IMHO, it's better to add this option if we really need this.
And if we need this right now, please leave a comment for this.

server/server.js
```diff
@@ -4,13 +4,20 @@
...
+app.use(bodyParser.urlencoded({
```
romandev: ditto.

PR #228, opened 2017-05-17
platform-test command added in gulpfile

Wuseok:


gulpfile.babel.js
```diff
@@ -77,3 +77,12 @@ gulp.task('stop', finish => {
     }, 1000);
   });
 });
+
+
```
romandev: Please remove unncessary line.

gulpfile.babel.js
```diff
@@ -77,3 +77,12 @@ gulp.task('stop', finish => {
...
+gulp.task('platform-test', function (cb) {
```
romandev: Please use underbar '_' instead of hyphen '-'.

romandev: Also, we should use arrow function instead of old style function as follows:
OLD style:
  function(cb) {
     ...
  }

NEW style:
  callback => {
    ...
  }

romandev: s/cb/finish

gulpfile.babel.js
```diff
@@ -77,3 +77,12 @@ gulp.task('stop', finish => {
...
+  child_process.exec('sh ./bootstrap/test/platform_test.sh', function (err, stdout, stderr) {
```
romandev: IMHO, we should implement this test in javascript. (and please remove the test files..)
When the gulp command was executed, |sync_node && set_path_env| would already have been executed.
Please see |absolute.sh| in details.

romandev:
any update?

gulpfile.babel.js
nadongguri: @romandev, would you explain again about implementing platform_test in javascript? :-)

PR #232, opened 2017-05-20
update npm when package.json is changed

nadongguri:
it solves [issue#218](https://github.com/romandev/absolute/issues/218)

### DIFF
--- a/bootstrap/absolute.sh
+++ b/bootstrap/absolute.sh
@@ -17,7 +17,7 @@ set_path_env $(bootstrap_command_path)
 set_path_env $(absolute_path)/node_modules/.bin
 
 # NPM update
-if [ ! .pkg_timestamp -nt package.json ]; then
+if [ ! -f .pkg_timestamp ] || [ package.json -nt .pkg_timestamp ]; then
   npm update && > .pkg_timestamp
 fi
 
--- a/bootstrap/absolute.sh
+++ b/bootstrap/absolute.sh
@@ -18,7 +18,7 @@ set_path_env $(absolute_path)/node_modules/.bin
 
 # NPM update
 if [ ! .pkg_timestamp -nt package.json ]; then
-  npm update && touch .pkg_timestamp
+  npm update && > .pkg_timestamp
 fi
 
 for command in $(ls $(bootstrap_command_path)); do
--- a/bootstrap/absolute.sh
+++ b/bootstrap/absolute.sh
@@ -17,7 +17,10 @@ set_path_env $(bootstrap_command_path)
 set_path_env $(absolute_path)/node_modules/.bin
 
 # NPM update
-npm update
+if [ ! .pkg_timestamp -nt package.json ]; then
+  npm update
+  touch .pkg_timestamp
+fi
 
 for command in $(ls $(bootstrap_command_path)); do
   if [ "$1" = "$command" ]; then
--- a/bootstrap/absolute.sh
+++ b/bootstrap/absolute.sh
@@ -18,8 +18,7 @@ set_path_env $(absolute_path)/node_modules/.bin
 
 # NPM update
 if [ ! .pkg_timestamp -nt package.json ]; then
-  npm update
-  touch .pkg_timestamp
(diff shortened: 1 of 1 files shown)


nadongguri:
One thing I'd like to check is after set path for node_modules/.bin, it includes touch command so, it prints out event though it acts like the same as touch in bash. I think @romandev  wouldn't like to print out of touch command so I'm thinking of removing that print.

bootstrap/absolute.sh
```diff
@@ -17,7 +17,10 @@ set_path_env $(bootstrap_command_path)
 set_path_env $(absolute_path)/node_modules/.bin
 
 # NPM update
-npm update
+if [ ! .pkg_timestamp -nt package.json ]; then
+  npm update
```
romandev: I think we should check whether |npm update| is success or not.
npm update && touch .pkg_timestamp

bootstrap/absolute.sh
```diff
@@ -17,7 +17,10 @@ set_path_env $(bootstrap_command_path)
 set_path_env $(absolute_path)/node_modules/.bin
 
 # NPM update
-npm update
+if [ ! .pkg_timestamp -nt package.json ]; then
```
romandev: |if [ package.json -nt .pkg_timestamp ]; then| is better?

romandev:
> One thing I'd like to check is after set path for node_modules/.bin, it includes touch command so,
> it prints out event though it acts like the same as touch in bash. I think @romandev wouldn't like
> to print out of touch command so I'm thinking of removing that print.

nadongguri@, I don't understand your opinion fully.
You think the |set_path_env| command also updates the modified time stamp like |touch| command.
So, you guess that I don't like the duplicated touch operation. Is my understanding correct?

bootstrap/absolute.sh
nadongguri: okay. I got the point

bootstrap/absolute.sh
nadongguri: Actually, I had to consider the case even .pkg_timestamp file is not created, so I wrote code as above.  |if [ package.json -nt .pkg_timestamp ]; goes to false when .pkg_timestamp file doesn't existed. :-)

nadongguri:
@romandev , code line 17 in /bootstrap/absolute.sh,
`set_path_env $(absolute_path)/node_modules/.bin`
afterthat, when I use touch command for updating timestamp to .pkg_timestamp file, it prints out like this.

root@dongguri:~/absolute_dongguri/nadongguri/absolute# ./absolute
.pkg_timestamp { dev: 64769,
  mode: 33188,
  nlink: 1,
  uid: 0,
  gid: 0,
  rdev: 0,
  blksize: 4096,
  ino: 786368,
  size: 0,
  blocks: 0,
  atime: 2017-05-21T02:17:50.000Z,
  mtime: 2017-05-21T02:17:50.000Z,
  ctime: 2017-05-21T02:17:50.341Z,
  birthtime: 2017-05-21T02:17:50.341Z }
[11:17:51] Requiring external module babel-register
[11:17:51] Using gulpfile ~/absolute_dongguri/nadongguri/absolute/gulpfile.babel.js
[11:17:51] Starting 'help'...
[11:17:51] Finished 'help' after 115 μs
[11:17:51] Starting 'default'...
[11:17:51] Finished 'default' after 15 μs
[11:17:51] Starting 'stop'...


I think script runs /node_modules/.bin/touch.
maybe everyone doesn't want to see that log, do they?

romandev:
@nadongguri Thank you for detailed explanation.

I think we can use redirection command '>' instead.
npm update && > .pkg_timestamp

WDYT?

nadongguri:
@romandev what a brilliant! I've commited. but many commits for a simple solution. :-)

bootstrap/absolute.sh
romandev: Oh, I agree with your opinion.
BTW, I did try what you said but it works well in my environment. (Mac)
Which platform did you try it on?

PR #233, opened 2017-05-21
change require() method to import for ES6,7 support

daehyunjimmy:
This patch is for issue#231
replacing require() to import for ECMA 6 and 7 supports.


romandev:
Overall looks good to me. Please update the database.js file as well.

Also, please update the commit description in more details.

PR #235, opened 2017-05-21
add help implementation by using gulp-help-doc

daehyunjimmy:
add help implementation using by gulp-help-doc

Usage: gulp [task] [options]
Tasks:


romandev:
@daehyunjimmy We are using gulp internal but users(some developers) will use our |absolute| command.

So, it should be displayed as "absolute" instead of "gulp".

PR #239, opened 2017-05-24
Add worker_manager for clustering

hyungheo:
It's about issue #241 (Node.js should use multi cores.)

worker_manager.js will make children processes for load-balancing by using cluster module.

### DIFF
--- a/server/worker_manager.js
+++ b/server/worker_manager.js
@@ -1,0 +1,29 @@
+// Copyright (c) 2017 The Absolute Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import cluster from 'cluster';
+import os from 'os';
+
+// Policy of dispatching task to another worker available
+// cluster.SCHED_NONE to leave it to the operating system
+// cluster.SCHED_RR for round-robin
+cluster.schedulingPolicy = cluster.SCHED_RR;
+
+if (cluster.isMaster) {
+  // master worker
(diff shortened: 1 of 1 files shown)


server/worker_manager.js
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2017 The Absolute Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import cluster from 'cluster';
+import os from 'os';
+
+cluster.schedulingPolicy = cluster.SCHED_RR;
```
romandev: What kind of schedulingPolicy is there?
Please add a comment with reference link here.

hyungheo: cluster.schedulingPolicy#
Added in: v0.11.2
The scheduling policy, either cluster.SCHED_RR for round-robin or cluster.SCHED_NONE to leave it to the operating system. This is a global setting and effectively frozen once either the first worker is spawned, or cluster.setupMaster() is called, whichever comes first.

SCHED_RR is the default on all operating systems except Windows. Windows will change to SCHED_RR once libuv is able to effectively distribute IOCP handles without incurring a large performance hit.

cluster.schedulingPolicy can also be set through the NODE_CLUSTER_SCHED_POLICY environment variable. Valid values are "rr" and "none".

ref.) https://nodejs.org/api/cluster.html

romandev: So, please add a comment in source code?

PR #244, opened 2017-05-25
bug-fix for supporting git-bash in windows

cs-lee:
1. add win-wget in third_party
2. export win-bash path for type command
3. sync third_party after set path


nadongguri:
You said you took a patch from Hackathon #151, so isn't set_path_env_last function needed to master?

cs-lee:
When exporting win-bash in the third_part of Absolute to path first, there was a problem that conflicts with commands of git-bash. So this is the function we need, But now the part where the problem occurred was deleted.

Issue #237, opened 2017-05-21
Add a lint rule not to allow the 'require' function to load module.


romandev:
To avoid confusing situation, it's better to restrict using the 'require' function. Let's add a new lint rule for it and then only use |import| instead of |require|.

nadongguri:
Okay.

daehyunjimmy:
when I try to find it with eslint config, it seems there is no rules for this restriction.

nadongguri:
@daehyunjimmy is right, there's no existing rule for disallowing require, so I will create a custom rule for it.

JayChl:
We can close this issue?

nadongguri:
I'm closing this issue. That feature is already applied in absolute 1.0.

PR #250, opened 2017-05-29
Enable eslint rule es6/no-var-requires

nadongguri:
Issue #237 is raised and to solve this issue, I created eslint-plugin-es6
repository then wrote eslint rule for forcing to use import instead of
require.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "babel-preset-es2015": "^6.0.15",
     "babel-preset-es2017": "^6.0.15",
     "eslint-config-google": "^0.7.1",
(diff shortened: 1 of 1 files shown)


romandev:
looks good to me but the following command doesn't work well..
eslint test.js --fix

nadongguri:
Okay I will check fix option.

nadongguri:
@romandev , I added fixer for no-var-requires rule.

romandev:
Please rebase this patch.

nadongguri:
@romandev, I think using eslint-plugin-import is better than this cuz it has only one simple thing. I will close this pull request for now. :-)

PR #251, opened 2017-05-30
Add push router and api

JayChl:
- server/{module}/api/router.js routing sub path
- add minimum required push apis


JayChl:
push apis and routers will commit after organizing it

PR #265, opened 2017-05-31
Fix build error

hyungheo:
Module("os") dosen't have a module("cpu").


romandev:
Please fix the root cause.

hyungheo:
It was run-time error.

PR #268, opened 2017-06-01
set babel cache path as $absolute_path

nadongguri:
when we use babel, cache file .babel.json is created in HOME path.
Absolute is designed to be portable so set BABEL_CACHE_PATH as
$absolute_path.


bootstrap/absolute.sh
```diff
@@ -26,6 +26,8 @@ if [ ! -f .pkg_timestamp ] || [ package.json -nt .pkg_timestamp ]; then
   npm update && > .pkg_timestamp
 fi
 
+export BABEL_CACHE_PATH=$(absolute_path)/.babel-cache.json
```
romandev: IMHO, this might not be a good way because the boostrap code(shell script) should not have a dependency with JS code if possible. The babel works in JS layer. So, we can set the environment before babel runs.

nadongguri: I agree with you. I will fond out the way to set babel path in js file.

nadongguri:
@romandev, I added TODO comment on that line but I guess there would be a graceful way to handle this issue.

bootstrap/absolute.sh
```diff
@@ -26,6 +26,9 @@ if [ ! -f .pkg_timestamp ] || [ package.json -nt .pkg_timestamp ]; then
   npm update && > .pkg_timestamp
 fi
 
+# TODO(nadongguri): this env vir should be set not here but before gulp command
```
romandev: Let's move these codes below bootstrap command loop

nadongguri: Okay!

nadongguri:
I moved line after loop. :-)

PR #273, opened 2017-06-02
Clean-up .gitignore file.

daehyunjimmy:
remove unnecessary items (ex. fcm)
add new items (ex. .pkg_timestamp and out)


romandev:
Why are you removing the fcm.js?
If you'd like to do it, please split this patch into Clean-up .gitignore and Remove fcm.js patches.
Or, you can update title and description in this patch without separated patches.

daehyunjimmy:
fcm is specific way for push, I want to change name and consider new design.
therefore, I will split this patch into Clean-up .gitignore and Remove fcm.js patches.

PR #274, opened 2017-06-02
Add |clean| command.

daehyunjimmy:
add clean commnad to remove directory
add test code to check clean


bootstrap/test/test-clean.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2017 The Absolute Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+const assert = require('assert');
+const execSync = require('child_process').execSync;
+
+describe('Clean Test', () => {
+  it('Is out directory cleaned?', () => {
+    //assert(fs.existsSync('./out/') == false);
```
romandev: We should add actual test codes.

bootstrap/test/test-clean.js
```diff
@@ -0,0 +1,13 @@
...
+    assert(true);
```
romandev: This is unnecessary?

gulpfile.babel.js
```diff
@@ -88,4 +88,10 @@ gulp.task('bootstrap_test', () => {
     .once('end', () => {
 	process.exit();	
     })
-});
\ No newline at end of file
+});
+
+gulp.task('clean', () => {
+  gulp.src(['out/'], {read: false})
```
romandev: I'd happy to add more things such as node_modules, .pkg_timestamp and so on.
If developer uses this command, should go back to initial status.

daehyunjimmy:
Now if I removes node_modules with this logic, there are some problems.
ex.) when process.on('exit'), it will call stop, but node_modules already removed. so it causes problem.
I will take care to fix that issue.

and when I start absolute bootstrap_test, exexSync not started well. (still checking)

bootstrap/test/test-clean.js
```diff
@@ -0,0 +1,17 @@
...
+// found in the LICENSE file.
+
+const assert = require('assert');
+const fs = require('fs');
+const path = require('path');
+const spawnSync = require('child_process').spawnSync;
+
+describe('Clean Test', () => {
+  it('Is out directory cleaned?', () => {
+    spawnSync(path.resolve('./absolute'), ['start']);
```
romandev: I think this command will block follow-up commands(stop and clean) because |start| command has a infinite-loop internally.
So, I think you can just execute |./absolute build| instead.
Then, the command will fetch all node modules and transpile all JS file into out directory.

gulpfile.babel.js
```diff
@@ -88,4 +89,11 @@ gulp.task('bootstrap_test', () => {
     .once('end', () => {
 	process.exit();	
     })
-});
\ No newline at end of file
+});
+
+// FIXME(jimmy): This command can not remove node_modules directory (need to fix it)
```
romandev: I'm not sure but this line looks longer than 80bytes?
If so, please add a line break.

bootstrap/test/test-clean.js
romandev: Let's improve this test.
1. Check whether |out| directory exists or not. (including  .pkg_timestamp) - ASSERT not exists
2. Run absolute build
3. Check whether |out| directory exists or not - ASSERT exists
4. Run absolute clean
5. Check whether |out| directory exists or not - ASSERT not exists again.

romandev:
This branch is out-of-date with the base branch.
Please rebase this branch.
Also, you can do test locally using the following command:
`./absolute bootstrap_test`

bootstrap/test/test-clean.js
daehyunjimmy: I tried with below way., but it seems that spawnSync for absolute cmd doesn't work.

    spawnSync(path.resolve('./absolute'), ['build']);
    assert(fs.existsSync('./out/') == true);
    spawnSync(path.resolve('./absolute'), ['clean']);
    assert(fs.existsSync('./out/') == false);

PR #276, opened 2017-06-04
Update README.md

nadongguri:
- Wrote form to explain the absolute project. 
- linked CI status


README.md
```diff
@@ -1,6 +1,20 @@
 Absolute
-=======
+=============
+
+### A web-based order platform like siren-order for Starbucks.
```
romandev: IMHO, this might be unnecessary because this sentence is duplicated with follow-up section.
WDYT?

README.md
```diff
@@ -1,6 +1,20 @@
...
+
+![Travis Build Status](https://travis-ci.org/romandev/absolute.svg?branch=master)
```
romandev: Can we add a chat icon with our slack link?

README.md
nadongguri: I'd like to present our project with one sentence actually. :-) I agree with you, many famous github projects explain what they are with one or short sentences so but not like us, I think we also need a good words to show us. anyway.

README.md
nadongguri: Okay. Not a problem.

README.md
nadongguri: WDYT about we need a short but powerful phrase explaining us? for now, I will remove duplicated one. :-)

romandev: Yeah, I respect your opinion. My point was about look-and-feel, not about content.
I did find some examples:
  - swift: https://github.com/apple/swift
  - linux: https://github.com/torvalds/linux
  - flutter: https://github.com/flutter/flutter
  - react: https://github.com/facebook/react

They have each their own style, but I didn't see a case that uses a |bold| style description below the project title. I thought that the consecutive |bold| style might be bad look-and-feel to present our project. But it's a personal taste. So, we can vote it. (for our members)

I agree with your opinion. We still need a short to present our project.
Also, if we have it, we can add it to the topic section (top of this page)

PR #277, opened 2017-06-04
 Webpack Enable in Absolute

cs-lee:
1. webpack package enable
2. Add the "webpack" command and set up webpack config.
3. Add the "webpack" sequence in "start" command
4. webpack watch mode on by default
5. entry file is "client/app.js"
6. output file is in client/dist


gulpfile.babel.js
```diff
@@ -83,9 +85,43 @@ gulp.task('bootstrap_test', () => {
...
     })
     .once('end', () => {
-	process.exit();	
+      process.exit();
     })
-});
\ No newline at end of file
+});
+
+gulp.task('webpack', () => {
```
romandev: Let's use |build_client| instead of |webpack|.
Also, existing |build| should be |build_server|.

client/index.html
```diff
@@ -3,6 +3,7 @@
 <head>
 </head>
 <body>
-Absolute Client!
+  <div id="root"></div>
+  <script src="dist/bundle.js"></script>
```
romandev: We should use out/client directory instead of dist.

gulpfile.babel.js
```diff
@@ -83,9 +85,43 @@ gulp.task('bootstrap_test', () => {
...
+    webpack({
+      watch: true,
+      context: path.resolve(__dirname, 'client'),
+      entry: './app.js',
+      output: {
+        path: path.resolve(__dirname, 'client', 'dist'),
```
romandev: Should be out/client

PR #278, opened 2017-06-04
Add support for Windows test bot using AppVeyor.

romandev:
This pull request is adding support for Windows test bot using AppVeyor.
Also, this CL is fixing bootstrap_test errors on Windows platform.

ISSUE=#139

Issue #139, opened 2017-03-01
Testing server for windows


hyungheo:
Travis is working for this project but it does not support the windows-os.

This project should support windows environment also.

So additional ci-service should be made for windows-os

romandev:
Looks great.

PR #279, opened 2017-06-05
Add support for CMD test on Windows.

romandev:
Introduced Windows test bot in #278 but it does test on PowerShell only.
But some developer might use a legacy CMD instead of PowserShell. So, we should
support for CMD test on Windows.

ISSUE=#139

### DIFF
--- a/absolute.cmd
+++ b/absolute.cmd
@@ -4,4 +4,5 @@
 
 @ECHO OFF
 set ABSOLUTE_PATH=%~dp0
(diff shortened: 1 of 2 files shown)


romandev:
This is unnecessary because our win bot is already running on CMD base.
We should add support for PowerShell instead of CMD.
I tried some ways but it doesn't work well yet.
So, I'll close this CL and then re-upload PowerShell patch again.

PR #280, opened 2017-06-05
Add support for PowerShell test on Windows.

romandev:
Introduced Windows test bot in #278 but it does test on CMD only. But some
developer might use PowerShell instead of the legacy CMD. So, we should
support for PowerShell test on Windows.

ISSUE=#139

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,6 +2,7 @@ environment:
   nodejs_version: "6"
 
 test_script:
+  - ps: ./absolute bootstrap_test 2>&1
   - ./absolute lint
   - ./absolute bootstrap_test
 
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   },
 
   "devDependencies": {
+    "minimatch": "^3.0.2",
     "babel-core": "^6.14.0",
     "babel-preset-es2015": "^6.0.15",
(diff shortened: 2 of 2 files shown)


nadongguri:
I have a question of this PR because "./absolute bootstrap_test" works in powershell with the latest of master branch. I'm wondering what patches mean. :-)

nadongguri:
I have another question of test failure of appveyor, will it be okay? :-)

./absolute : npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
At line:1 char:1
+ ./absolute bootstrap_test 2>&1
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (npm WARN deprec...egExp DoS issue:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError

romandev:
Thank you for your interesting. This is not perfect yet.
The AppVeyor is currently using CMD as default shell instead of PowerShell.
So, I had to add |ps:| keyword to the |test_script:| section.
BTW, in the PowerShell, if printing something out to stderr, then AppVeyor recognizes that error occurs.
So, we should fix the problem first. I think we can do workaround by disabling stderr but I'm not sure if it is a good way.

Issue #270, opened 2017-06-02
bootstrap/test/test-update.js implementation


Wuseok:
We need to check 'npm upate' working or not.
Current status is 
If package.json file is modified, to check modified date of this file, start to npm update command.
after finish update, it saved update time to .pkg_timestamp file. 

We have to check 
if(timestamp > package.json modified time) true : false

romandev:
Can you implement this test before next meeting?

Wuseok:
Hello romandev I update pull request. 
Could you check it?

nadongguri:
@Wuseok, I'm closing this issue.

PR #282, opened 2017-06-06
Check npm update module was added in bootstrap test

Wuseok:
Following issue #270
I added to check npm update module.

I try to test Windows and Linux.
Please check it

### DIFF
--- a/bootstrap/test/test-update.js
+++ b/bootstrap/test/test-update.js
@@ -11,6 +11,11 @@ const path = require('path');
 const fs = require('fs');
 
 describe('Update Test', () => {
+
+  it('Is npm update check file exist?', () => {
+	assert(fs.existsSync(".pkg_timestamp"));
+  });
+
   it('Is npm update working?', () => {
 	var pkg_stats = fs.statSync("package.json");
 	var timestamp_stats = fs.statSync(".pkg_timestamp");
--- a/bootstrap/test/test-update.js
+++ b/bootstrap/test/test-update.js
@@ -7,12 +7,14 @@
 // please check github issue number #242
 
 const assert = require('assert');
-const execSync = require('child_process').execSync;
+const path = require('path');
+const fs = require('fs');
 
 describe('Update Test', () => {
-  it('Is update was fine?', () => {
-	// FIXME(Wuseok): It will be updated new test module here.
-	// test-update.js will be check about npm update is successfully or not.
-	assert(true);
+  it('Is npm update working?', () => {
+	var pkg_stats = fs.statSync("package.json");
+	var timestamp_stats = fs.statSync(".pkg_timestamp");
+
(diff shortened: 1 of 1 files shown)


bootstrap/test/test-update.js
```diff
@@ -7,12 +7,14 @@
...
+const path = require('path');
+const fs = require('fs');
 
 describe('Update Test', () => {
-  it('Is update was fine?', () => {
-	// FIXME(Wuseok): It will be updated new test module here.
-	// test-update.js will be check about npm update is successfully or not.
-	assert(true);
+  it('Is npm update working?', () => {
+	var pkg_stats = fs.statSync("package.json");
```
romandev: I think you should do this test as follows:
* Test case 1. If the .pkg_timestamp doesn't exist, then .pkg_timestamp should be created after updating.
* Test case 2. If the .pkg_timestamp exists, then .pkg_timestamp should be modified (timestamp) after updating.

Wuseok: Ok. I will add to check timestamp file is created or not.

bootstrap/test/test-update.js
```diff
@@ -7,12 +7,19 @@
...
+const fs = require('fs');
 
 describe('Update Test', () => {
-  it('Is update was fine?', () => {
-	// FIXME(Wuseok): It will be updated new test module here.
-	// test-update.js will be check about npm update is successfully or not.
-	assert(true);
+
+  it('Is npm update check file exist?', () => {
+	assert(fs.existsSync(".pkg_timestamp"));
```
romandev: This is not a right way. I'll explain about my previous comment in more details.
> Test case 1. If the .pkg_timestamp doesn't exist, then .pkg_timestamp should be created after updating.

It means that if the .pkg_timestamp doesn't exist, then `absolute` command have to execute `npm update`.
So, the test should look like this:
1. assert .pkg_timestamp not exists
2. run absolute command
3. (Optionally) if possible, it's better to check whether some packages are actually installed or not.
4. assert .pkg_timestamp exists (if absolute command is not failed)

> Test case 2. If the .pkg_timestamp exists, then .pkg_timestamp should be modified (timestamp) after updating.

We probably can split this test into two small tests.
One thing should look like this:
1. assert .pkg_timestamp exists
2. assert .pkg_timestamp's modified time is newer than package.json.
3. Keep .pkg_timestamp's modified time to use it in step 5.
4. run absolute command
5. assert .pkg_timestamp's modified time == kept modified time (It means that the .pkg_timestamp's modified time is not updated)

Another thing should look like this:
1. assert .pkg_timestamp exists
2. assert package.json's modified time is newer than .pkg_timestamp's modified time.
3. Keep .pkg_timestamp's modified time to use it in step 5.
4. run absolute command
5. assert .pkg_timestamp's modified time is newer than kept modified time and package.json's modified time (It means that the .pkg_timestamp's modified time should be updated)

PR #283, opened 2017-06-06
Database Connection

coconutperm:
It need connect to database server when started server.


server/server.js
```diff
@@ -9,6 +9,7 @@ import bodyParser from 'body-parser';
 import config from './config';
 import httpsServer from './https_server/https_server';
 import redirectServer from './https_server/redirect_server';
+import dbConnect from './db/db_connect';
```
romandev: Do we really need this wrapper?
I think you can just use mongoose.connect() directly.
Also, please use promise version instead of old style version.
http://mongoosejs.com/docs/connections.html

server/db/db_connect.js
```diff
@@ -0,0 +1,27 @@
...
+function connectServer() {
+  mongoose.connection.on('error', console.error);
+  mongoose.connection.once('open', () => {
+    console.log('Connected to mongod server');
+  });
+  mongoose.connection.on('close', () => {
+    console.log('Mongoose disconnected');
+  });
+
+  mongoose.connect('localhost:27017/absolute', (error) => {
```
romandev: I think the URL is defined as named constant in other common file.

server/db/db_connect.js
```diff
@@ -0,0 +1,27 @@
...
+    console.log('mongodb connected');
+    if (error) {
+      console.error('mongodb connection error', error);
+    }
+  });
+}
+
+exports.connectServer = connectServer;
```
romandev: We should use ecma6 export.
For example, we can do it as follows:
```
export function hello() {
    // This function is exported.
}
```

Issue #287, opened 2017-06-06
How client get vapid public key for subscribtion


JayChl:
To encrypt push payload, we use vapid which has private / public key
Public key need to know client side but it's different for each server instance.
[refer web-push readme](https://github.com/web-push-libs/web-push#using-vapid-key-for-applicationserverkey)

Suggest opening api to get public vapid key

JayChl:
#349 implement push vapid public key for client side to create data file.

Issue #286, opened 2017-06-06
Generating server vapid keys for first execution


JayChl:
To send payload, need to encrypt payload data using vapid which is need to generate once(will not be removed after released)

JayChl:
Resolved #318

Issue #285, opened 2017-06-06
Testing push notification


JayChl:
Testing Push notification needs
- Running on some browsers (chrome, samsung browser, safari, etc)
- Known endpoint and subscription keys(auth, p256dh)
- 365-day connected device to web and server to update the result after getting push event

nadongguri:
@JayChl, , I'm closing this issue, you can re-open it if you want.

PR #288, opened 2017-06-06
Add sendPush function using web-push for payload encryption

JayChl:
- Need to generate vapid for first execution #286
- Need to get server vapid public key #287 
- Need to make testcase #285

### DIFF
--- a/server/push/controllers/send_push.js
+++ b/server/push/controllers/send_push.js
@@ -5,8 +5,6 @@
 import config from '../../config';
 import webpush from 'web-push';
 
-const serverSubject = 'https://github.com/romandev/absolute';
-
 /**
  * Send push with payload to endpoint with it's authentication key.
  * @param {string} endpoint endpoint from subscription
@@ -15,10 +13,11 @@ const serverSubject = 'https://github.com/romandev/absolute';
  * @return {promise} result includes statusCode, headers, body
  */
 export function sendPush(endpoint, keys, payload) {
+  const SERVER_SUBJECT = 'https://github.com/romandev/absolute';
   return new Promise((resolve, reject) => {
     webpush.setGCMAPIKey(config.serverInfo.pushServerKey);
     webpush.setVapidDetails(
-      serverSubject,
+      SERVER_SUBJECT,
       config.serverInfo.pushVapidKeys.publicKey,
       config.serverInfo.pushVapidKeys.privateKey,
     );
--- a/server/push/controllers/send_push.js
+++ b/server/push/controllers/send_push.js
@@ -1,0 +1,42 @@
+// Copyright (c) 2017 The Absolute Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import config from '../../config';
+import webpush from 'web-push';
+
+/**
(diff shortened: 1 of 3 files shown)


server/push/controllers/send_push.js
```diff
@@ -0,0 +1,42 @@
...
+import webpush from 'web-push';
+
+/**
+ * Send push with payload to endpoint with it's authentication key.
+ * @param {string} endpoint endpoint from subscription
+ * @param {string} keys authentication key and p256dh key from subscription
+ * @param {string} payload payload data send to endpoint
+ * @return {promise} result includes statusCode, headers, body
+ */
+function sendPush(endpoint, keys, payload) {
```
romandev: I think you can use |export| in ECMA6 feature as follows:
```
export function sendPush(endpoint, keys, payload) {
    ...
}
```

server/push/controllers/send_push.js
```diff
@@ -0,0 +1,42 @@
...
+    webpush.sendNotification(pushSubscription, payload)
+    .then((webPushResult) => {
+      return resolve(webPushResult);
+    }).catch((webPushError) => {
+      return reject(webPushError);
+    });
+});
+}
+
+module.exports = sendPush;
```
romandev: If you use export function, then this is unnecessary.

server/push/controllers/send_push.js
JayChl: Done! Thanks for the information

server/push/controllers/send_push.js
```diff
@@ -0,0 +1,42 @@
...
+ * @param {string} endpoint endpoint from subscription
+ * @param {string} keys authentication key and p256dh key from subscription
+ * @param {string} payload payload data send to endpoint
+ * @return {promise} result includes statusCode, headers, body
+ */
+function sendPush(endpoint, keys, payload) {
+  return new Promise((resolve, reject) => {
+    webpush.setGCMAPIKey(config.serverInfo.pushServerKey);
+    webpush.setVapidDetails(
+      'https://github.com/romandev/absolute',
```
JayChl: I tried, but web-push throws error for here
  if (typeof subject !== 'string' || subject.length === 0) {
    throw new Error('The subject value must be a string containing a URL or ' +
      'mailto: address. ' + subject);
  }
So, use variable for the parameter to readability

server/push/controllers/send_push.js
JayChl: Done

server/push/controllers/send_push.js
romandev: Thank you for detailed explanation!

server/push/controllers/send_push.js
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2017 The Absolute Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import config from '../../config';
+import webpush from 'web-push';
+
+const serverSubject = 'https://github.com/romandev/absolute';
```
romandev: Let's move this subject into function and I think this should be uppercase only?
const SERVER_SUBJECT = ...
like this?

Issue #300, opened 2017-06-11
Source-map is not applied in server


hyungheo:
When node make some error or exception , origin-source code is not traced by node.

hyungheo:
619730c014d8add064394a8583dd24d16331679d

PR #301, opened 2017-06-11
Apply sourcemap to server (about Issue #300)

hyungheo:
This patch provides source map support for stack traces in node.

It uses the source-map module to replace the paths and line numbers of source-mapped files with their original paths and line numbers. The output mimics node's stack trace format with the goal of making every compile-to-JS language more of a first-class citizen. Source maps are completely general (not specific to any one language) so you can use source maps with multiple compile-to-JS languages in the same node process.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,8 +26,9 @@
     "run-sequence": "^1.0.1",
     "sass-loader": "^6.0.5",
     "style-loader": "^0.18.2",
-    "url-loader": "^0.5.8",
-    "webpack": "^2.6.1"
+    "source-map-support": "^0.4.15",
+    "webpack": "^2.6.1",
+    "url-loader": "^0.5.8"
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -20,7 +20,8 @@
     "gulp-sourcemaps": "^1.6.0",
     "mocha": "^3.4.1",
     "run-sequence": "^1.0.1",
-    "webpack": "^2.6.1"
+    "webpack": "^2.6.1",
+    "source-map-support": "^0.4.15"
```
cs-lee: We need to sort alphabetical order.

hyungheo: It has been fixed

romandev:
Please rebase this

hyungheo:
rebasing has been done

PR #313, opened 2017-06-13
fix absolute start error

nadongguri:
Latest master branch occurs error when running "absolute start".
It is because of not existing database folder so handled it.


nadongguri:
One thing I'm worrying about is the code mkdir -p in gulp file. Will it be better do it in absolute script file?

romandev:
@nadongguri Thank you for this patch.

If you don't mind, let's use a easier way.
1. Make database directory.
2. Put just .gitkeep file (its content should be empty)
3. Then, there is no code like mkdir -p blah blah in gulp file.

nadongguri:
Okay. I appreciate your kindness guide. :-)

nadongguri:
Made it out!

Wuseok:
oh... I'm late.

PR #315, opened 2017-06-14
add workbox-webpack-plugin to install service-worker

daehyunjimmy:
* workbox-sw
workbox-sw is a high-level wrapper that ties together the smaller modules into a single, consistent interface.
For most developers, workbox-sw will be all your service worker needs to handle routing and caching.
workbox-sw handles both precaching (using a manifest generated by workbox-cli, workbox-webpack-plugin, or workbox-build), and runtime caching.

to generate service worker, add workbox-webpack-plugin.


daehyunjimmy:
I am checking using workbox-build instead of this.
(https://workboxjs.org/reference-docs/latest/module-workbox-build.html#.generateSW)

daehyunjimmy:
https://github.com/GoogleChrome/workbox/issues/605

PR #318, opened 2017-06-15
Add generating push vapid key

JayChl:
- Move push related key to push_key.json
- Generate push vapid key at first launching server


JayChl:
need fix result path out json

JayChl:
Reading/Writing target json path is "../../server/push/push_key.json".
Adding json file to babel build can make server stop unexpectedly
#323

Issue #297, opened 2017-06-09
To create Server test environment


Wuseok:
In case of server side test, we should to start server before testing.
I will create test folder like this (/server/test/) and putting start server module before start testing.

Wuseok:
Hello @romandev 
I checked to start server in mocha but it is not easy.
I have some issue and failed to start absolute server.

1. start command (./absolute start) using spawnSync
- I try to start server using this command but ENOENT error was comes up.
2. /server/test/ is not possible because this code < return gulp.src(['./server/**/*.js']) > this command is included gulpfile.babel.js
3. Access out/server folder and start server.js file
- If bootstrap test was succeed, out/server folder and file was created but is it OK about this flow?

romandev:
I'll take this.

nadongguri:
@Wuseok, I'm closing this issue.

PR #321, opened 2017-06-15
To create Server test environment

Wuseok:
I mentioned Issue #297
In case of server side test, we should to start server before testing.

Here I added 'server_test' command in gulpfile.js
and after './absolute start' command passed, /test/*.js file will be started.
but here has some issue.
* /server/test/ is not possible because below code 
< return gulp.src(['./server/**/*.js']) > [gulpfile.babel.js]
this command was built with test-*.js file so build error was occurred.
so temporary I create test folder in parent folder.

Please review this.


romandev:
Leave a comment

Issue #325, opened 2017-06-20
"build_client" does not terminate on Ctrl+C


cs-lee:
"build_client" does not terminate on Ctrl+C

nadongguri:
@cs-lee, I'm closing this issue, you can re-open it if you want. :D

PR #326, opened 2017-06-20
bug fix - "build_client" does not terminate on Ctrl+C #325

cs-lee:
1. Add the signal handler in build_client sequence.

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -19,6 +19,9 @@ process.on('exit', () => {
 
 process.on('SIGINT', () => {
   runSequence('stop');
+	process.nextTick( () => {
+	  process.exit(1);
+	});
 });
 
 process.on('uncaughtException', () => {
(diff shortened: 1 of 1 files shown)


gulpfile.babel.js
```diff
@@ -139,4 +139,8 @@ gulp.task('build_client', () => {
   }, (err, stats) => {
       // FIXME(cs-lee) save log in file
   });
+
+  process.on('SIGINT', () => {
```
romandev: This event is already defined at top block of this file.

romandev: I'm not sure if this is a right way.
Can you ensure that the other daemon or service is stopped safely?

nadongguri:
Pressing "Ctrl + C" could end build_client task during running
using the exist SIGINT handler. But I'm afraid of using nested
process function calls.

PR #333, opened 2017-06-25
Migration product api

coconutperm:


server/order/router/product_router.js
```diff
@@ -0,0 +1,28 @@
...
+});
+
+router.get('/', (request, response) => {
+  product.getProductList()
+    .then((productList) => {
+      response.send(productList);
+    });
+});
+
+module.exports = router;
```
romandev: We don't use module.exports. Please use ECMA version export.

server/order/router/product_router.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2017 The Absolute Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import express from 'express';
+import * as product from '../controller/product_controller';
+
+const router = express.router();
+
+router.post('/', (request, response) => {
```
romandev: Let's make a new style router.

server/order/router/product.router.js
```diff
@@ -0,0 +1,37 @@
...
+  get(request, response) {
+    product.getProductList()
+      .then((productList) => {
+        response.send(productList);
+      });
+  }
+
+  post(request, response) {
+    if (request.body.name && request.body.amount) {
+      product.addProduct(JSON.parse(JSON.stringify(request.body)))
```
romandev: Why are you calling stringify() and then parse() again?

server/order/controller/product_controller.js
```diff
@@ -0,0 +1,40 @@
...
+import {Product} from '../model/product';
+
+/**
+ * Add product data what admin added.
+ * @param {json} productData data what admin sent.
+ * @return {promise} result product data what admin sent.
+ */
+export function addProduct(productData) {
+  return new Promise((resolve, reject) => {
+    Product.findOneAndUpdate({name: productData.name}
```
romandev: As far as I know, the mongoose already supports Promise as follows:
  - http://mongoosejs.com/docs/promises.html

server/order/controller/product_controller.js
```diff
@@ -0,0 +1,40 @@
...
+  });
+}
+
+/**
+ * Get product data list to saved.
+ * @return {promise} result data list of product
+ */
+export function getProductList() {
+  return new Promise((resolve, reject) => {
+    Product.find({}, (error, productList) => {
```
romandev: ditto

server/order/controller/product_controller.js
```diff
@@ -0,0 +1,40 @@
...
+      if (productList) {
+        resolve(JSON.stringify(productList));
+      } else {
+        reject(-1);
```
romandev: This is not right way.
We should define Error object with error message explicitly.

server/order/router/product.router.js
```diff
@@ -0,0 +1,37 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Application from '../../base/application.js';
+import * as product from '../controller/product_controller';
+
+@Application.route('/product')
```
romandev: Add jsdoc-style comments.

server/order/router/product.router.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  post(request, response) {
+    if (request.body.name && request.body.amount) {
+      product.addProduct(JSON.parse(JSON.stringify(request.body)))
+        .then((product) => {
+          response.send(product);
+        });
+    } else {
+      response.sendStatus(400);
```
romandev: Please use named constant.

coconutperm: The error code must be formally defined.

server/order/controller/product_controller.js
coconutperm: Mongoose queries are not promises. However, they do have a .then() function for yield and async/await. If you need a fully-fledged promise, use the .exec() function.

server/order/controller/product_controller.js
coconutperm: Mongoose queries are not promises. However, they do have a .then() function for yield and async/await. If you need a fully-fledged promise, use the .exec() function.

romandev: Yeah, but you can just use then() or exec() instead.
IMHO, wrapper seems unnecessary, do you agree with me?

PR #338, opened 2017-06-25
Add service worker and service worker manager

daehyunjimmy:
- Add service worker (copy from client/service-worker.js to out/client/)
- Add service worker manager to register
- Remove unnecessary file.


client/service-worker.js
```diff
@@ -0,0 +1,9 @@
+importScripts(
```
romandev: This looks unnecessary?

gulpfile.babel.js
```diff
@@ -106,6 +106,8 @@ gulp.task('bootstrap_test', () => {
 gulp.task('build_client', () => {
   gulp.src(path.resolve(__dirname, 'client', 'index.html'))
```
romandev: This gulp.src() can take array instead of calling two times :)

PR #341, opened 2017-06-28
Add POST /api/push to send push notification

JayChl:
- add push router
- add absolute push server key (this key will be deprecated after released)
- fix getting push key from sendPush function


JayChl:
Added push server key for absolute created new firebase project for testing

JayChl:
parent commit has changes, raise new pr

PR #344, opened 2017-07-02
Add POST /api/push to send push notification

JayChl:
- add push router
- add absolute push server key (this key will be deprecated after released)
- fix getting push key from sendPush function

### DIFF
--- a/server/push/controller/send_push.js
+++ b/server/push/controller/send_push.js
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-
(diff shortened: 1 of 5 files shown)


server/push/controller/send_push.js
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import config from '../../config';
+
```
romandev: Please remove unnecessary line

JayChl: Done

JayChl:
Need to resolve conflict with new router decorator

PR #345, opened 2017-07-02
Implement Application class.

romandev:
The Application class is a wrapper class to hide server backend implementation.
In our Absolute project, we are currently using the Express but might want to
change the backend to other things. Even if so, we can change them smoothly.
This change is introducing a new way to route requests. We can refer to the
related example in server/example/router/example.router.js.

### DIFF
--- a/server/base/application.js
+++ b/server/base/application.js
@@ -1,0 +1,89 @@
+// Copyright (c) 2017 The Absolute Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


server/base/application.js
```diff
@@ -0,0 +1,88 @@
...
+   * @param {String} path
+   */
+  start(path) {
+    requireAll({
+      dirname: path,
+      filter: /\.router\.js$/,
+      recursive: true,
+    });
+
+    this._app.listen(3000);
```
JayChl: Do we use port 3000 instead of 9080 or 9443?
And how can we use https?

romandev: Done.

After this patch, we block the https. The previous way is not good to me.
I'll fix it in follow-up CL.

PR #351, opened 2017-07-08
Router refactoring

coconutperm:
We need to implement new router style.


server/authorization/router/auth.router.js
```diff
@@ -12,29 +12,28 @@
...
-const router = express.router();
-
-router.post('/uuid', (request, response) => {
-  auth.generateUuid().then((authInfo) => {
-    response.send(authInfo);
-  }, (error) => {
-    response.sendStatus(400);
-  });
-});
+@Application.route('/uuid')
```
romandev: Please add a jsdoc comments for rest API.
  - example: https://github.com/romandev/absolute/pull/344/files

server/authorization/router/auth.router.js
```diff
@@ -12,29 +12,28 @@
...
+export default class UuidRouter {
```
romandev: We are using UUID here but also we are using UserID in other place.
I think UserID(or UserKey) is better than UUID. What do you think?

server/authorization/router/auth.router.js
```diff
@@ -12,29 +12,28 @@
...
+  get(request, response) {
+    if (request.query.uuid) {
+      auth.getUuid(request.query.uuid).then((authInfo) => {
+        response.json(authInfo);
+      }, (error) => {
+        response.sendStatus(400);
```
romandev: I think the 400 error is incorrect here.
The 400 error means "bad request" but it happens when the url syntax error.
So, we probably should use 403 error "Forbidden" instead of 400.

romandev: Also, I think we should use named constant instead of this number. (for readability)
Is there no named constant pre-defined in Express or Node?

server/authorization/router/auth.router.js
```diff
@@ -12,29 +12,28 @@
...
 
-router.get('/uuid', (request, response) => {
-  if (request.query.uuid) {
-    auth.getUuid(request.query.uuid).then((authInfo) => {
-      response.json(authInfo);
+  post(request, response) {
+    auth.generateUuid().then((authInfo) => {
+      response.send(authInfo);
     }, (error) => {
       response.sendStatus(400);
```
romandev: ditto

server/authorization/router/auth.router.js
```diff
@@ -12,29 +12,28 @@
...
+      response.sendStatus(400);
+    }
+  }
 
-router.get('/uuid', (request, response) => {
-  if (request.query.uuid) {
-    auth.getUuid(request.query.uuid).then((authInfo) => {
-      response.json(authInfo);
+  post(request, response) {
+    auth.generateUuid().then((authInfo) => {
```
romandev: ditto (Please use UserID or other thing)

server/authorization/router/auth.router.js
```diff
@@ -12,29 +12,28 @@
...
+        response.sendStatus(400);
+      });
+    } else {
+      response.sendStatus(400);
+    }
+  }
 
-router.get('/uuid', (request, response) => {
-  if (request.query.uuid) {
-    auth.getUuid(request.query.uuid).then((authInfo) => {
```
romandev: ditto (Please use UserID or other thing)

server/authorization/router/auth.router.js
```diff
@@ -12,29 +12,28 @@
...
+      }, (error) => {
+        response.sendStatus(400);
+      });
+    } else {
+      response.sendStatus(400);
+    }
+  }
 
-router.get('/uuid', (request, response) => {
-  if (request.query.uuid) {
```
romandev: (Please use UserID or other thing)

server/authorization/router/auth.router.js
```diff
@@ -12,29 +12,28 @@
...
+        response.json(authInfo);
+      }, (error) => {
+        response.sendStatus(400);
+      });
+    } else {
+      response.sendStatus(400);
+    }
+  }
 
-router.get('/uuid', (request, response) => {
```
romandev: (Please use UserID or other thing)

server/authorization/router/auth.router.js
```diff
@@ -12,29 +12,28 @@
...
-  auth.generateUuid().then((authInfo) => {
-    response.send(authInfo);
-  }, (error) => {
-    response.sendStatus(400);
-  });
-});
+@Application.route('/uuid')
+export default class UuidRouter {
+  get(request, response) {
+    if (request.query.uuid) {
```
romandev: (Please use UserID or other thing)

server/authorization/router/auth.router.js
```diff
@@ -12,29 +12,28 @@
...
+      auth.getUuid(request.query.uuid).then((authInfo) => {
```
romandev: (Please use UserID or other thing)

server/authorization/router/auth.router.js
```diff
@@ -12,29 +12,28 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import express from 'express';
+import Application from '../../base/application.js';
 import * as auth from '../controller/auth_controller';
 
-const router = express.router();
-
-router.post('/uuid', (request, response) => {
```
romandev: (Please use UserID or other thing)

server/authorization/router/auth.router.js
```diff
@@ -12,29 +12,28 @@
...
-  auth.generateUuid().then((authInfo) => {
```
romandev: (Please use UserID or other thing)

server/authorization/router/auth.router.js
```diff
@@ -12,29 +12,28 @@
...
-    auth.getUuid(request.query.uuid).then((authInfo) => {
-      response.json(authInfo);
+  post(request, response) {
+    auth.generateUuid().then((authInfo) => {
+      response.send(authInfo);
     }, (error) => {
       response.sendStatus(400);
     });
-  } else {
-    response.sendStatus(400);
```
romandev: ditto

server/order/router/order.router.js
```diff
@@ -12,25 +12,24 @@
...
 
-const router = express.router();
-
-router.post('/', (request, response) => {
-  if (request.body.userId) {
-    order.saveOrder(JSON.parse(JSON.stringify(request.body)));
-    response.sendStatus(200);
-  } else {
-    response.sendStatus(400);
+@Application.route('/order')
```
romandev: Please add jsdoc-style comments.

server/order/router/order.router.js
```diff
@@ -12,25 +12,24 @@
...
+export default class OrderRouter {
+  get(request, response) {
+    order.getOrderList(request.query.userId)
```
romandev: I'm not sure if this is right way.
Once we authorize user connection, then we keep the session.
Okay, we should do in the past but didn't
Therefore, let's leave a TODO or FIXME comment at least.

server/order/router/order.router.js
```diff
@@ -12,25 +12,24 @@
...
-    .then((orderList) => {
-      response.send(orderList);
-    });
-});
 
-module.exports = router;
+  post(request, response) {
+    if (request.body.userId) {
+      order.saveOrder(JSON.parse(JSON.stringify(request.body)));
+      response.sendStatus(200);
```
romandev: Please use named constant.

server/order/router/order.router.js
```diff
@@ -12,25 +12,24 @@
...
+    } else {
+      response.sendStatus(400);
```
romandev: ditto

server/authorization/router/auth.router.js
romandev: I tried to find pre-defined constant value but not exists.
So, please ignore my previous comment.

romandev: I think this should be 403 error because the userKey is for authorization.
But I'm not sure. @JayChl, What do you think?

JayChl: I agree with 403  :)

Issue #357, opened 2017-07-10
push_key.js files are always tracked after starting absolute.


romandev:
The two push_key.js files(c/p/push_keys.js, s/p/push_keys.js) are modified after starting absolute because creating a new key for push. It causes inconvenient.

So, what should we do? One way is that we add their names to .gitignore.

romandev:
@JayChl, @daehyunjimmy Can you follow this issue up?

JayChl:
Will add this to .gitignore

JayChl:
It's better to preserve server/push/push_key.js's form to make code simply,
however .gitignore is not affect to already commited file. 
commands
$ git update-index --assume-unchanged server/push/push_key.js
$ git update-index --assume-unchanged server/push/push_key.js; 
achive this purpose. Can we add this command to script?

JayChl:
Closing this issue #372

PR #360, opened 2017-07-10
Add push_key.js to .gitignore #357

JayChl:
- push vapid keys are newly generated and added
    by starting server to push_key.js so file changes.
- add client/push/push_key.js to .gitignore not to commit
- add 'git update-index --assume-unchanged server/push/push_key.js' after generating push keys

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -86,6 +86,12 @@ gulp.task('lint_router', finish => {
 
 gulp.task('push_key', () => {
   generatePushKey();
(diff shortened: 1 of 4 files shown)


gulpfile.babel.js
```diff
@@ -86,6 +86,12 @@ gulp.task('lint_router', finish => {
 
 gulp.task('push_key', () => {
   generatePushKey();
+  child_process.exec('git update-index --assume-unchanged'
```
romandev: I think it would be enough to run this command onetime.

PR #364, opened 2017-07-11
add push class to control push notification

daehyunjimmy:
make push class to control push logic


client/push/Push.js
```diff
@@ -0,0 +1,19 @@
+export default class Push {
```
romandev: We should add boilerplate at top of this file.

client/push/Push.js
```diff
@@ -0,0 +1,19 @@
...
+  constructor() {
+    /* Token to use for push notification */
+    this._token = 0;
+  }
+
+  addPushToServer(token) {
```
romandev: You should add jsdoc-style comments. (all exported functions and classes)

client/push/push_manager.js
```diff
@@ -0,0 +1,41 @@
...
+  constructor() {
+    /* Token to use for push notification */
+    this._token = 0;
+  }
+
+  /**
+   * send push information to server
+   */
+  addPushToServer(token) {
+    this._token = token;
```
JayChl: This could be adding push subscription to server, how about registerSubscription?
eg) push.registerSubscription(subscription)

client/push/push_manager.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  /**
+   * get push information from server
+   */
+  getPushFromServer() {
```
JayChl: What will do at getPushFromServer?  it is getting registered push subscription from server using uuid?

daehyunjimmy: yes I made this to get push subscription with uuid from server.
I will change this name to getSubscription.

PR #365, opened 2017-07-11
Add PushRouter using Application.route

JayChl:
- add router for push using Application.route
- use push_key instead of config

### DIFF
--- a/server/push/router/push.router.js
+++ b/server/push/router/push.router.js
@@ -1,6 +1,16 @@
-// Copyright (c) 2017 The Absolute Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE file.
+// Copyright (c) 2017 The Absolute Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 
 import Application from '../../base/application.js';
 import {sendPush} from '../controller/send_push';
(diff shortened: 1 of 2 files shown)


server/push/router/push.router.js
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2017 The Absolute Authors. All rights reserved.
```
romandev: We should replace this with a new thing.
Please see example.router.js

server/push/router/push.router.js
```diff
@@ -0,0 +1,32 @@
...
+    if (request.body.endpoint && request.body.keys) {
+      sendPush(request.body.endpoint, request.body.keys,
+        JSON.stringify(request.body.payload))
+        .then((result) => {
+          response.send(result);
+        }).catch((error) => {
+          response.send(error);
+        });
+    } else {
+      response.sendStatus(400);
```
romandev: I think the 400 error is incorrect here.
The 400 error means "bad request" but it happens when the url syntax error.
So, we probably should use 403 error "Forbidden" instead of 400. (but I'm not sure)

JayChl: IMHO, From newly updated RFC, 400 Bad Request can apply to here
>    The 400 (Bad Request) status code indicates that the server cannot or
   will not process the request due to something that is perceived to be
   a client error (e.g., malformed request syntax, invalid request
   message framing, or deceptive request routing).

on the other hands,
> The 403 (Forbidden) status code indicates that the server understood
   the request but refuses to authorize it.  A server that wishes to
   make public why the request has been forbidden can describe that
   reason in the response payload (if any).
 https://tools.ietf.org/html/rfc7231#page-58

So in this case we can apply malformed request syntax here

server/push/router/push.router.js
JayChl: done thanks!

server/push/router/push.router.js
```diff
@@ -0,0 +1,42 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Application from '../../base/application.js';
+import {sendPush} from '../controller/send_push';
+
+@Application.route('/api/push/')
```
romandev: nit: /api/push (Please remove last slash.)

server/push/router/push.router.js
romandev: Acknowledgement!

romandev:
Latest patch looks good to me but please defer merging this patch for now because Travis and AppVeyor are not working well...

PR #379, opened 2017-07-23
Add @JayChl, @coconutperm, @nadongguri to CODEOWNERS.

romandev:
There will be a lot of pull requests in each modules but we don't have enough
owners for review process. So, we make the following people become temporary
owners to digest each module reviews. They are not official owners who are
nominated by official process but are much contributing to each modules.
So, we believe that they can take responsibility for each modules.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,18 @@
 # Absolute Project Owners
 * @romandev
 
-# Absolute Sub-module Owners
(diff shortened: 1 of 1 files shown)


romandev:
PTAL @coconutperm, @JayChl, @nadongguri 
I added you to OWNER list.

romandev:
I'll wait for other people's comments.

daehyunjimmy:
Good.

PR #383, opened 2017-07-26
Add api /api/push/token for managing client push token

JayChl:
- POST : add or update push tokens using userId to server
- GET : get push tokens using userId from server

push tokens includes endpoint, p256dh, auth key


server/push/router/push.token.router.js
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) 2017 The Absolute Authors.
```
romandev: I think this filename should be push_token.router.js

PR #384, opened 2017-07-26
import push module in service worker

daehyunjimmy:
- modify webpack entry & output
- import push module in service worker


client/service-worker.js
```diff
@@ -1,3 +1,5 @@
+importScripts('javascript/sw.js');
```
romandev: Let's use ECMA import instead of this.
For example,
`import Push from './push/push_manager.js';`

gulpfile.babel.js
```diff
@@ -136,10 +136,13 @@ gulp.task('build_client', () => {
   webpack({
     watch: true,
     context: path.resolve(__dirname, 'client'),
-    entry: './app.js',
+    entry: {
+      bundle : './app.js',
+      sw : './push/push_manager.js'
```
romandev: Replace this line with `sw : './service-worker.js'`
Also, you should remove 134 line.

client/service-worker-manager.js
```diff
@@ -1,3 +1,5 @@
+import Push from './push/push_manager.js';
+
 if ('serviceWorker' in navigator) {
   window.addEventListener('load', function() {
     navigator.serviceWorker.register('service-worker.js').then(
```
romandev: Should be `navigator.serviceWorker.register('javascript/sw.js').then(`

PR #385, opened 2017-07-27
Apply the ESlint rule to 'gulpfile.babel.js'

GarlicB:
I fix various ESLint errors in the file. 
(camelcase, semicolon, comma dangle, brace style, indentation)

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -13,9 +13,9 @@
 // limitations under the License.
 
 import babel from 'gulp-babel';
-import child_process from 'child_process';
+import childProcess from 'child_process';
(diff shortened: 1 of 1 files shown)


romandev:
Just curious: What tool did you use? Also, our lint test is not checking for gulpfile.babel.js?

GarlicB:
Yes, 'gulpfile.babel.js' did not check for itself. because it checked only certain .js files in 'server' folder. 
so i added './gulpfile.babel.js' on gulp.src (task of 'lint_server').

And I checked this file with ESlinter of the Atom at first.
Actually. I think the result is the same as checking with the gulp-eslint module in our gulpfile.babel.js.

PR #387, opened 2017-07-30
handle push notification event in service worker

daehyunjimmy:
* create Push notification when service worker got push event
* handle onclick event listener on push notification

### DIFF
--- a/client/push/notification_manager.js
+++ b/client/push/notification_manager.js
@@ -57,8 +57,12 @@ export default class Notification {
    * process notification click event
    */
   processNotificationClickEvent(event) {
-    let url = '';
-	event.waitUntil(clients.openWindow(url));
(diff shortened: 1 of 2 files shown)


client/push/notification_manager.js
```diff
@@ -28,13 +28,37 @@ export default class Notification {
...
-  closeNotification() {
+  closeNotification(event) {
+    event.notification.close();
+  }
+
+  /**
+   * process notification click event
+   */
+  processNotificationClickEvent(event) {
+    let url = '';
```
romandev: Why is this url empty string?

daehyunjimmy: I changed it using event.notification.data
I will add getting information from server if no data and handling event action logic.

PR #389, opened 2017-08-03
Custom command for pr

GarlicB:
- Written in JS. 
- It's located in /bootstrap/command/cc.js.
- To work with the gulp task, edited gulpfile.babel.js (adding task 'cc', 'cp').
- Added the gulp option '--silent' in absolute.sh.

Use the 'readline' module for multiple inputs, But I think this part is quite rough. 
(This is why it took a long time from the last time.)

Always appreciate your review.


gulpfile.babel.js
```diff
@@ -139,17 +141,16 @@ gulp.task('bootstrap_test', () => {
...
     context: path.resolve(__dirname, 'client'),
-    entry: {
-      bundle: './app.js',
-      sw: './service-worker.js'},
+    entry: './app.js',
     output: {
-      path: path.resolve(__dirname, 'out', 'client'),
-      filename: '[name].js'},
+      path: path.resolve(__dirname, 'out', 'client', 'javascript'),
+      filename: 'bundle.js'},
```
daehyunjimmy: - please check duplicated app.js entry
- need to keep [name].js filenama and path to generate right service worker file

GarlicB: thanks, i'll check.

bootstrap/absolute.sh
```diff
@@ -47,4 +47,9 @@ done
 # TODO(nadongguri): this env vir should be set not here but before gulp command
 export BABEL_CACHE_PATH=$(absolute_path)/.babel-cache.json
 
-gulp $@
+# ccl requires several steps. it need to turn off the gulp message
+if [ "$@" = "cc" ]; then
+  gulp $@ --silent
```
nadongguri: As far as I heard, putting silent option is to avoid async call sequence. But is there other ways to handle?

GarlicB: Easy to say, the reason I put the '--silent' option on the 'cc' task is that the gulp message is printed every time when typing an answer. I'll see another way.

nadongguri:
Would you remove .babel-cache.json?

nadongguri:
I don't know why but import {inputRL, rl} from './bootstrap/command/cc'  blocks ending of gulp process.

GarlicB:
Close this, because my request have some problem about running project. 
I will summarize my thoughts and write new issue.

PR #391, opened 2017-08-05
register push subscription information to server

daehyunjimmy:
add fetch request function


client/common/util_manager.js
```diff
@@ -0,0 +1,51 @@
...
+    const SERVER_URL = `//${HOST}:${PORT}`;
+    return SERVER_URL;
+  }
+
+  /**
+   * fetch request operation
+   */
+  static fetchRequest(targetUrl, method, data) {
+    return new Promise(function (resolve, reject) {
+      fetch(targetUrl, { method, body: data }).then(function (response) {
```
nadongguri: There should be no space after '{' and '}' by ECMA lint. :-)

client/common/util_manager.js
```diff
@@ -0,0 +1,51 @@
...
+    const PORT = self.location.port;
+    const SERVER_URL = `//${HOST}:${PORT}`;
+    return SERVER_URL;
+  }
+
+  /**
+   * fetch request operation
+   */
+  static fetchRequest(targetUrl, method, data) {
+    return new Promise(function (resolve, reject) {
```
nadongguri: Unexpected space before function parentheses by ECMA lint, so every line wrote "function (" should be "function("

client/common/util_manager.js
daehyunjimmy: Thanks I modified that :)

client/common/util_manager.js
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2017 The Absolute Authors.
```
JayChl: util_manager seems strange, how about just util.js?

client/push/push_manager.js
```diff
@@ -30,12 +31,38 @@ export default class Push {
...
   registerSubscription(subscription) {
     this._subscription = subscription;
+    const SERVER_URL = Util.getServerURL();
+    const queryUrl = `${SERVER_URL}/api/push/token/`;
+    let pushHeaders = new Headers({
+      'Content-Type': 'application/json'
+    });
+    let pushData = JSON.stringify({
+      // FIXME(daehyun): this userId should be replaced by real user id
+      userId: '1234',
```
JayChl: Let's use subscription.endpoint for userId temporary.
if we use fixed userId, it will always overwrite userId-subscription that already registered to server

client/common/util_manager.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  /**
+   * fetch request operation
+   */
+  static fetchRequest(request) {
+    return new Promise(function(resolve, reject) {
+      fetch(request).then(function(response) {
+        if (response.status !== 200) {
+          console.log('Failed to fetch from ${targetUrl} Status Code: '
```
JayChl: 'Failed to fetch from ${targetUrl} Status Code: '  should be
`Failed to fetch from ${targetUrl} Status Code: ` for use targetUrl as variable

client/push/push_manager.js
```diff
@@ -30,6 +31,31 @@ export default class Push {
    */
   registerSubscription(subscription) {
     this._subscription = subscription;
+    const SERVER_URL = Util.getServerURL();
+    const queryUrl = `${SERVER_URL}/api/push/token/`;
```
romandev: Do we need |absolute URL|? Can't we just use |relative URL|?

client/common/util.js
```diff
@@ -0,0 +1,48 @@
...
+    const SERVER_URL = `//${HOST}:${PORT}`;
+    return SERVER_URL;
+  }
+
+  /**
+   * fetch request operation
+   */
+  static fetchRequest(request) {
+    return new Promise(function(resolve, reject) {
+      fetch(request).then(function(response) {
```
romandev: The fetch() API is already promise-based API[1]. Why do we need this wrapper?

[1] https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Basic_concepts

client/common/util.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+export default class Util {
+  /**
+   * get server URL
+   */
+  static getServerURL() {
+    const HOST = self.location.hostname;
+    const PORT = self.location.port;
+    const SERVER_URL = `//${HOST}:${PORT}`;
+    return SERVER_URL;
```
romandev: We can simplify this function as follows:
return `//${self.location.hostname}:${self.location.port}`

PR #393, opened 2017-08-06
Ignore babel-cache.json

GarlicB:
Add '.babel-cache.json' in '.gitignore'

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,8 @@
 *.swp
+.babel-cache.json
 .pkg_timestamp
 node_modules
 out
 third_party/mongodb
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ out
 third_party/mongodb
 third_party/node
 database
+.babel-cache.json
```
romandev: We should keep a alphabetical order.
So, this change should be at line number 2. (before .pkg_timestamp)

GarlicB: Oh..ok i just fixed it according to the rules.

PR #398, opened 2017-08-15
Initial codes for typescript.

romandev:
This patch imports TypeScript modules and then introduce gulpfile.ts to replace
gulpfile.babel.js. Until finishing this migration, we can use the following
command to run gulpfile.ts.
  $ ./absolute --gulpfile gulpfile.ts <command>


nadongguri:
Checked it late. :D

PR #412, opened 2017-08-20
Add @daehyunjimmy to CODEOWNERS.

romandev:
There will be a lot of pull requests in each modules but we don't have enough
owners for review process. So, we make the following people become temporary
owners to digest each module reviews. They are not official owners who are
nominated by official process but are much contributing to each modules.
So, we believe that they can take responsibility for each modules.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -9,9 +9,9 @@
 # nominated by official process but are much contributing to each modules.
 # So, we believe that they can take responsibility for each modules.
 # Push
-server/push/* @romandev @JayChl
-client/push/* @romandev @JayChl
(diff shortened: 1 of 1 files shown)


romandev:
Hi @nadongguri and @JayChl,

I'd like to nominate @daehyunjimmy to push's owner.
He is actively working on Push module.

@nadongguri WDYT as a project leader?
@JayChl WDYT as another push owner?

JayChl:
I absolutely agree!

romandev:
@nadongguri Gentle ping

nadongguri:
I am late, what will I say? I totally agree. :D

PR #418, opened 2017-08-30
typescript apply in client build sequence

cs-lee:
After this patch, app.ts will build by typescirpt compiler and webpack so bundle_type.js will create in out path.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "gulp-mocha": "^4.3.1",
     "gulp-nodemon": "^2.2.1",
     "gulp-sourcemaps": "^1.6.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -28,6 +28,7 @@
     "gulp-mocha": "^4.3.1",
     "gulp-nodemon": "^2.2.1",
     "gulp-sourcemaps": "^1.6.0",
+    "gulp-typescript": "^3.2.2",
```
romandev: Where do we use this module?

cs-lee: This module does not use in this patch, it will use in next patch which for "typescript apply in server"

PR #419, opened 2017-08-30
add server test using mocha

JayChl:
- Enable server test for travis and appveyor
- Add some push test cases


server/push/test/test-push.js
```diff
@@ -0,0 +1,100 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import assert from 'assert';
+import {describe, it} from 'mocha';
+import * as pushSender from '../controller/send_push';
+
```
romandev: Please remove unnecessary line :)

server/push/test/test-push.js
```diff
@@ -0,0 +1,100 @@
...
+
+console.log('Welcome Absolute Push Test');
```
romandev: Isn't it better to remove this console log?

server/push/test/test-push.js
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) 2017 The Absolute Authors.
```
romandev: IMHO, push_test.js is better. Let's discuss about this tomorrow.

server/push/test/test-push.js
```diff
@@ -0,0 +1,100 @@
...
+  title: 'Absolute Test Payload',
+  body: 'Push Test body :)',
+};
+
+describe('[POSITIVE] Push Send', () => {
+  it('Send push with fixed push tokens', (done) => {
+    pushSender.sendPush(TEST_PUSH_TOKENS.endpoint, TEST_PUSH_TOKENS.keys,
+      JSON.stringify(TEST_PAYLOAD))
+      .then((result) => {
+        assert.equal(result.statusCode, 201);
```
romandev: Why is this 201? Shouldn't be 200?

server/push/test/test-push.js
```diff
@@ -0,0 +1,100 @@
...
+        done();
+      }).catch((error) => {
+        console.log('Failed to sendPush!' + error);
```
romandev: I think returning promise way is better?
https://wietse.loves.engineering/testing-promises-with-mocha-90df8b7d2e35
http://programmingsummaries.tistory.com/383

JayChl:
Change returning promise instead of using done() for testing async functions

server/push/test/test_push.js
```diff
@@ -0,0 +1,133 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import assert from 'assert';
+import {describe, it} from 'mocha';
+import * as pushSender from '../controller/send_push';
+
+const TEST_PUSH_TOKENS_PC_CHROME = {
+  endpoint: 'https://fcm.googleapis.com/fcm/send/fvkYiDvVn_k:APA91bH8h4uZK4F' +
```
romandev: IMHO, we should implement mock-push service(?) if possible instead of actual service.
(Because the system is down or changed or something happens I think)
But I think this test is good for now. Let's improve it incrementally.

server/push/test/test_push.js
```diff
@@ -0,0 +1,133 @@
...
+      }).catch((error) => {
+        assert.equal(error.statusCode, 400);
+      });
+  });
+
+  it('Send push with invalid P256dh key', () => {
+    return pushSender.sendPush(TEST_PUSH_TOKENS_PC_CHROME.endpoint,
+      INVALID_P256_KEYS, JSON.stringify(TEST_PAYLOAD))
+      .then((result) => {
+        assert.notEqual(result.statusCode, 201);
```
romandev: I'm not sure but this shouldn't happen, right?
If so, shouldn't we assert error here?

server/push/test/test_push.js
JayChl: Yes right, but we need to check positive case.
So my plan is register push token as valid (to get 201 from push service) one. but not actual used for developing this project and not used for real user 
(for PC browser, push is not recieved when browser is not launched or connected to server.
for mobile browser, device will be old one that is not used anymore)

server/push/test/test_push.js
JayChl: yes, it shouldn't happen.  So it should not get 201(normal case) and normally it is ok catching error.
(browsers can have different behavior. some browser does not catching error)

PR #442, opened 2017-09-10
add indexed db to save push User Id

daehyunjimmy:
- open Indexed DB 
- save push User ID to DB
- need implement to save more values


client/common/db_manager.js
```diff
@@ -0,0 +1,113 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * DataBase class (indexed DB)
+ */
+const DB_NAME = 'Absolute';
+export default class DataBase {
```
romandev: Usually prefer to use Database rather than DataBase.

romandev:
Sorry for delayed review.

PR #446, opened 2017-09-21
update vue_client

Wuseok:


Wuseok:
Don't review this source code now.
It is not regular update.
checking for competition with Heo hyung.
We have to merge UI source code but first we check together about this.

PR #474, opened 2017-10-02
Separate configuration for TypeScript from gulpfile.ts

romandev:
Currently, the configuration is hardcoded in gulpfile.ts. It's not bad
but one problem is that the configuration doesn't apply to gulpfile.ts.

ISSUE=none

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -50,21 +50,14 @@ gulp.task('test', ['build_server']);
 gulp.task('run_server', () => {
   nodemon({
     script: 'out/server.js',
-    ignore: 'out/'
+    ignore: ['out/']
   });
 });
 
+const tsProject = tsc.createProject('tsconfig.json');
 gulp.task('build_server', () => {
-  return gulp.src('./server/**/*.ts')
-    .pipe(tsc({
-      target: 'es5',
-      noImplicitReturns: true,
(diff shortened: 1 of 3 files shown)


gulpfile.ts
```diff
@@ -50,21 +50,14 @@ gulp.task('test', ['build_server']);
 gulp.task('run_server', () => {
   nodemon({
     script: 'out/server.js',
-    ignore: 'out/'
+    ignore: ['out/']
```
yjaeseok: I think [] block we do not need now.
How do you think add [] block after we need multiple arguments?

romandev: Thank you for review.
In Javascript, we could do it because there is no explicit types.
But after this patch, applying strong-type checking, the ```ignore``` parameter should be string[] type as per definition.
You can see the definition here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/gulp-nodemon/index.d.ts#L39

romandev: @yjaeseok I have to wait for your reply but I don't have enough time now. So, I'll merge this patch and then review this again (TBR)

yjaeseok: Oh sooorrry !

romandev:
Thank you for review

PR #475, opened 2017-10-02
Introduce jest for module unit tests.

romandev:
In Absolute v2, one of the most important things is unit test. So, we
introduce jest for module unit tests. After this patch, we can run test
as follows:
  $ ./absolute test

Also, this change is adding an example test in server/example/.

ISSUE=none

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,16 +3,40 @@
     "@types/express": "^4.0.37",
     "@types/gulp": "^3.8.33",
     "@types/gulp-nodemon": "0.0.30",
+    "@types/gulp-shell": "0.0.30",
+    "@types/jest": "^21.1.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -3,16 +3,40 @@
...
+    "supertest": "^3.0.0",
+    "ts-jest": "^21.0.1",
     "ts-node": "^3.3.0",
     "tslint": "^5.7.0",
     "tslint-microsoft-contrib": "^5.0.1",
     "typescript": "^2.5.3"
+  },
+  "jest": {
+    "moduleFileExtensions": [
+      "ts",
```
yjaeseok: how do you think change ordering by alphabet?

romandev:
Thank you for review. Another look?

Issue #467, opened 2017-10-02
Fix lint errors in server directory.


romandev:
We introduced tslint in https://github.com/lunchclass/absolute/commit/430af4a7892ec3593dff8bf19858fb008953b385. But there are still many lint errors. So, we should fix the lint errors manually as per the guide message. FYI, we are using [Micrsoft TypeScript Style Guide](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines).

You can check the lint errors as follows:
```
$ ./absolute lint
```

If you feel difficult to resolve the errors, please feel free to leave some questions here.
Also, you can split this change into multiple patches.

yjaeseok:
I will try this issue !!
but too many issue we have, I would like to do with together :)

haeungun:
I'll take this one together :D

PR #478, opened 2017-10-03
Fix Typescript Lint error basic.

yjaeseok:
I fix typescript lint error.

ISSUE=#467

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -1,3 +1,6 @@
 {
-  "extends": "tslint-microsoft-contrib"
+  "extends": "tslint-microsoft-contrib",
+  "rules": {
+	  "no-relative-imports": false
+  }
 }
--- a/tslint.json
+++ b/tslint.json
@@ -1,3 +1,3 @@
 {
-  "extends": "tslint-microsoft-contrib",
+  "extends": "tslint-microsoft-contrib"
 }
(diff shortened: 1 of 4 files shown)


yjaeseok:
I request it, but we have more lint error in server folder!

tslint.json
```diff
@@ -1,3 +1,8 @@
 {
-  "extends": "tslint-microsoft-contrib"
+  "extends": "tslint-microsoft-contrib",
+  "rules": {
+	  "no-suspicious-comment": false,
+	  "no-reserved-keywords": false,
+	  "no-relative-imports": false
```
romandev: We should fix the issue instead of suppressing this rule. WDYT?

yjaeseok: Your means that we fix all lint issue?

yjaeseok: @romandev Okay, I remove this ignore rules. :(

romandev:
@yjaeseok, could you update commit description? You seem to remove that ignore cases?
(Patch looks good to me.)

tslint.json
romandev: I don't understand your previous comments exactly :)
I wanted to fix all lint errors in the future (not in this patch).
If you thought that some rules are unnecessary, we can discuss about it together! (in Slack or here)

yjaeseok: Okay, once the ignore rule has been removed,
After that, it would be better to check if there are any unnecessary rules.

yjaeseok:
@romandev Okay, I updated commit description.
Next, we should discuss about ignore rules.

PR #481, opened 2017-10-05
Fix lint errors in server directory

haeungun:
Fixed the function names and variable names in the server directory.

ISSUE=#467

### DIFF
--- a/server/base/application.ts
+++ b/server/base/application.ts
@@ -30,7 +30,7 @@ export class Application {
     return this.app;
   }
 
-  public static route(url: string) {
+  public static ROUTE(url: string) {
     const app: express.Application = this.app;
 
     return <T>(routerClass: { new(...args: {}[]): T}): void => {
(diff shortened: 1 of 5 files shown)


server/base/application.ts
```diff
@@ -16,22 +16,22 @@
 
 import * as express from 'express';
 
-export default class Application {
-  private static app_: express.Application = express();
+export class Application {
+  private static app: express.Application = express();
 
-  public static async start() {
+  public static async START() {
```
romandev: Thank you for this change!
BTW, it seems a little bit weird to me.
Could you explain why this function should be upper case name? (including startForTesting and route)
Is that also a kind of lint error? If so, we might have to disable the rule.

haeungun: I changed it to uppercase because the lint error caused the static method name to be in upper case. (lint option: function-name)

romandev: Thank you for detailed explanation.
I'll re-check it again after merging this patch :)

Issue #471, opened 2017-10-02
Implement ./absolute clean command


romandev:
If we use the command, then remove all built outputs in out/ directory.
You can add a new task in gulpfile.ts as follows and then implement it simply.
```TypeScript
gulp.task('clean', (callback) => {
  // remove out/ directory
});
```
You can see the reference here: https://github.com/gulpjs/gulp/blob/master/docs/API.md

sapzape:
I'll take this issue

nadongguri:
Close this issue, #487

PR #482, opened 2017-10-05
Add ./absolute clean command

sapzape:
I added a command named clean.
This command deletes all outputs in out/ directory.

ISSUE=#471


gulpfile.ts
```diff
@@ -55,6 +56,10 @@ gulp.task('run_server', () => {
   });
 });
 
+gulp.task('clean', (callback) => {
```
nadongguri: I think callback argument is probably unnecessary in this case.

gulpfile.ts
```diff
@@ -55,6 +56,10 @@ gulp.task('run_server', () => {
...
+    return del('out/*', {force: true});
```
romandev: I think it's better to use 'out' instead of 'out/*'.
We want to remove the out directory as well as out/* files(and sub-directories).

Also, could you write a test for this patch?
You can write the test in bootstrap/test/clean.test.ts. (The example test code is in ./server/example/example.test.ts)

You might use the following code snippet:
```typescript
child_process.spwanSync('./absolute', ['build_server']);   // This will create out/ directory
fs.existsSync('./out');   // You can check whether the out directory exists
child_process.spawnSync('./absolute', ['clean']);    // This will remove out/ directory
fs.existsSync('./out');   // You can check whether the out directory exists
```

gulpfile.ts
sapzape: Okay. I remove callback argument in this code.
Thank you for review :D

gulpfile.ts
sapzape: Okay. I modified argument value to remove the out directory (+ files and sub-directories).
I also added a test code file named clean.test.ts to bootstrap/test/.
Thank you for review :D

PR #487, opened 2017-10-07
Implement ./absolute clean command

sapzape:
I added a command named clean.
This command deletes all outputs in out/ directory.
I also added test about clean command.

ISSUE=#471


bootstrap/test/clean.test.ts
```diff
@@ -0,0 +1,11 @@
+import {} from 'jest';
+import * as childProcess from 'child_process';
+import * as fs from 'fs';
+
+test('GET /test', async() => {
+    childProcess.spawnSync('./absolute', ['build_server']);
```
romandev: Could you try childProcess.spawnSync('./absolute', ['build_server'], { shell: true });?

bootstrap/test/clean.test.ts
```diff
@@ -0,0 +1,11 @@
...
+    expect(fs.existsSync('./out')).toBe(true);
+
+    childProcess.spawnSync('./absolute', ['clean']);
```
romandev: ditto.

romandev:
@nadongguri Could you check why this is broken in Windows?
I guess the problem might be path problem coming from different file system.
The following similar test worked well in Absolute v1. Could you please check it with ```path```?
https://github.com/lunchclass/absolute/blob/absolute_v1/bootstrap/test/test_update.js#L46

(I tried but my Windows system is not working well now..)

nadongguri:
@romandev, I will check it.

bootstrap/test/clean.test.ts
```diff
@@ -0,0 +1,11 @@
+import {} from 'jest';
+import * as childProcess from 'child_process';
+import * as fs from 'fs';
+
+test('GET /test', async() => {
+    childProcess.spawnSync('./absolute', ['build_server'], { shell: true });
```
nadongguri: that line doesn't work properly on Windows targets, so run ./absolute command as
childProcess.spawnSync(path.resolve('./absolute'), ['build_server'], { shell: true });

bootstrap/test/clean.test.ts
```diff
@@ -0,0 +1,11 @@
+import {} from 'jest';
+import * as childProcess from 'child_process';
+import * as fs from 'fs';
+
```
nadongguri: Add import *as path from 'path';

bootstrap/test/clean.test.ts
```diff
@@ -0,0 +1,11 @@
...
+test('GET /test', async() => {
+    childProcess.spawnSync('./absolute', ['build_server'], { shell: true });
+    expect(fs.existsSync('./out')).toBe(true);
+
+    childProcess.spawnSync('./absolute', ['clean'], { shell: true });
```
nadongguri: Add path.resolve the same as described above.

sapzape:
When i modified code 'childProcess.spawnSync('absolute', ['build_server'], { shell: true });' then, it work.
I think if the OS is Windows, then it seems to use 'absolute' as a parameter.

Issue #469, opened 2017-10-02
Add manifest.json in client side to support Web App Manifest.


romandev:
The web app manifest provides information about an application (such as name, author, icon, and description) in a JSON text file. The purpose of the manifest is to install web applications to the homescreen of a device, providing users with quicker access and a richer experience. Web app manifests are part of a collection of web technologies called progressive web apps, which are web applications that can be installed to the homescreen of a device without needing the user to go through an app store, along with other capabilities such as  being available offline and receiving push notifications.

Please see: https://developers.google.com/web/fundamentals/web-app-manifest/?hl=ko

elliekwon:
I'll take this issue

PR #488, opened 2017-10-07
Add client path and web app manifest

elliekwon:
- I added client path and index.html / manifest.json files.
- Manifest info can check through below url after run server :
`http://127.0.0.1:8090/manifest.json`
- But currently manifest has minimal information only 
`eg. no launcher image, no gcm_sender_id...`
- Web app will not installed yet because that without service worker. 

ISSUE=#469

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -18,8 +18,4 @@
 <head>
   <link rel="manifest" href="./manifest.json">
 </head>
-<body>
-  <div id="root"></div>
-  <script src="bundle_type.js"></script>
-</body>
 </html>
--- a/client/index.html
+++ b/client/index.html
@@ -1,0 +1,25 @@
+<!DOCTYPE html>
+<!--
+ Copyright (c) 2017 The Absolute Authors.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http:www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


client/index.html
```diff
@@ -0,0 +1,25 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+--!>
+<html>
+<head>
+  <link rel="manifest" href="./manifest.json">
+</head>
+<body>
+  <div id="root"></div>
+  <script src="bundle_type.js"></script>
```
romandev: Please remove this line (because there is no bundle_type.js file.)

server/base/application.test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import {} from 'jest';
+import * as supertest from 'supertest';
+import {Application} from '../base/application';
+
+test('GET /', async() => {
+  const request: {} = supertest(await Application.START_FOR_TESTING());
+  const response: {} = await request.get('/');
+  expect(response.statusCode).toBe(200);
+  expect(response.text).toBe('hello world');
```
romandev: Please remove this.

server/base/application.ts
```diff
@@ -15,12 +15,14 @@
...
 
 import * as express from 'express';
+import * as path from 'path';
 
 export class Application {
   private static app: express.Application = express();
 
   public static async START(): Promise<void> {
     await import('../example/example.router');
+    this.app.use(express.static(path.join(__dirname, '../../client')));
```
romandev: I think you should add this line to START_FOR_TESTING().

Issue #468, opened 2017-10-02
Add some badges in README.md.


romandev:
We have already two badges. Travis and AppVeyor. But we still need some additional badges such as Slack link and License link.

So, you can just update the following links in README.md:
  - https://img.shields.io/badge/slack-online-green.svg
  - https://img.shields.io/badge/License-Apache%202.0-yellowgreen.svg

onyoon7:
I want to take this issue : )

nadongguri:
@onyoon7, how's it going?

GarlicB:
Hi, I'll take this one together.

PR #497, opened 2017-10-23
Add badges in README.md

GarlicB:
Add two badges to our README.md file.

ISSUE=#468


nadongguri:
LGTM, one trival thing, would you change the order of the slack badge and the license badge to the license badge and the slack badge the same as the bacardi proejct?

GarlicB:
OK

PR #501, opened 2017-10-29
Implement skeleton codes for Notification APIs

sphilee:


nadongguri:
PTAL @JayChl, @daehyunjimmy for the notification design.
@sphilee, please fill the description. :)

client/apis/notification.ts
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+export default class Notification {
+    async close(): Promise<boolean> {
+        // Not implemented yet
+        return false;
+    }
+
+    async requestPermission(): Promise<boolean> {
```
romandev: Do we need this function?
(I know that this is needed for original API but we should hide all things if possible.)

romandev: Could you write some examples?

PR #502, opened 2017-10-31
Implement skeleton codes for Notification APIs

sphilee:
This is an initial implementation to provide notification APIs to
client app. The APIs always return false right now and we can implement
them later.

ISSUE=none

### DIFF
--- a/client/notification/notification.test.ts
+++ b/client/notification/notification.test.ts
@@ -18,11 +18,14 @@ import { } from 'jest';
 import absolute from '../absolute';
 
 test('absolute.notification.create()', async () => {
-  expect(await absolute.notification.create(event)).toBe(false);
+  let NotificationEvent  = new Event("click");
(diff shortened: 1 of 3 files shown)


client/notification/notification.test.ts
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { } from 'jest';
+import absolute from '../absolute';
+
+test('absolute.notification.create()', async () => {
+  expect(await absolute.notification.create(event)).toBe(false);
```
romandev: Where is ```event``` defined at?
If this line, 24 line, and 27 line will make bot errors.

sphilee:
It's hard to understand because the notification API is not friendly with me.

sphilee:
@romandev It was fixed :)

Issue #503, opened 2017-10-31
Add new gulp task to 'gulpfile.js' for running mongo daemon when Absolute service start.


llighter:
I'm going to apply mongo daemon startup option to 'gulpfile.ts' for running mongodb when Absolute service start.

PR #505, opened 2017-10-31

llighter:
ISSUE= #503

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -19,13 +19,13 @@ import * as gulp from 'gulp';
 import * as nodemon from 'gulp-nodemon';
 import * as runSequence from 'run-sequence';
 import tslint from 'gulp-tslint';
-import * as shell from 'gulp-shell';
 import * as tsc from 'gulp-typescript';
+import * as process from 'child_process';
 
 const webpack = require('webpack-stream');
 
 gulp.task('default', (callback) => {
-  runSequence('webpack', 'build_server', 'run_server', callback);
+  runSequence('webpack', 'build_server', 'run_server', 'start-mongo', callback);
 });
 
 gulp.task('lint', () => {
@@ -49,7 +49,7 @@ gulp.task('lint:fix', () => {
     }));
 });
 
-gulp.task('test', ['webpack'], shell.task('jest'));
+gulp.task('test', ['webpack'], runCommand(`jest`));
 
 gulp.task('run_server', () => {
   nodemon({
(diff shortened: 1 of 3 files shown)


gulpfile.ts
```diff
@@ -75,3 +76,17 @@ gulp.task('build_server', () => {
     .js
     .pipe(gulp.dest('out/'));
 });
+
+const mongodb_path = './third_party/mongodb';
+gulp.task('start-mongo', runCommand(`${mongodb_path}/bin/mongod --dbpath ${mongodb_path}`));
```
romandev: We are already using gulp.shell. So, can we just use gulp.shell() instead of runCommand()?

package.json
```diff
@@ -9,6 +9,7 @@
     "@types/request": "^2.0.3",
     "@types/run-sequence": "0.0.29",
     "@types/supertest": "^2.0.3",
+    "child_process": "^1.0.2",
```
romandev: Even If we have to child_processs, do we really need this? The child_process modules is built-in library as far as I know. WDYT?

gulpfile.ts
llighter: Oh, I didn't know about 'gulp.shell'. I will look into it and fix it.

package.json
llighter: If we can use the built-in library, it's a good idea to use it. I'll look into it too. :)

gulpfile.ts
romandev: @llighter I had to give you the information early but I couldn't. I'm sorry for that. Also, thank you for this work!

llighter: @romandev It's fine. I was able to study this in more detail.

package.json
romandev: But if you use gulp.shell, we might not need it :)

gulpfile.ts
llighter: @romandev I've found that now gulp do not recommend 'gulp.shell' anymore.
Here is [blackList.json file]( https://github.com/gulpjs/plugins/blob/master/src/blackList.json) at gulpjs repo.
Gulp recommend to use [gulp-exec]() or [child_process]().
And according to [gulp-exec]() docs, If you just want to run a command, don't use this plugin.
And [gulp API docs](https://github.com/gulpjs/gulp/blob/master/docs/API.md#gulpdestpath-options) use [child_process](https://github.com/gulpjs/gulp/blob/master/docs/API.md#gulpdestpath-options) like this.
```
// run a command in a shell
var exec = require('child_process').exec;
gulp.task('jekyll', function(cb) {
  // build Jekyll
  exec('jekyll build', function(err) {
    if (err) return cb(err); // return error
    cb(); // finished task
  });
});
```
What should I use?
WDYT? PTAL :)

romandev: Okay! Thank you for the information. If so, we can move this patch to gulp-exec or child_process.
And, once this patch is merged, you can also replace gulp-shell with gulp-exec or child_process in existing codes.

llighter: @romandev ok, then I will replace gulp-shell with child_process in our code.

llighter:

PR #508, opened 2017-11-02
[WIP] Implement register for Push API

haeungun:


client/push/push_manager.ts
```diff
@@ -19,25 +19,46 @@ declare var navigator: any;
...
+
+          if (!registration.pushManager) {
+            return false;
+          }
+
+          registration.pushManager.subscribe({
+            userVisibleOnly: true
+          })
+          .then((subscription: PushSubscription) => {
+            console.log(subscription);
```
romandev: Please remove console.log.

client/push/push_manager.ts
```diff
@@ -19,25 +19,46 @@ declare var navigator: any;
...
       navigator.serviceWorker.register('/push_service_worker.js')
         .then((registration: ServiceWorkerRegistration) => {
-          this.isRegistered_ = true;
+
+          if (!registration.pushManager) {
+            return false;
+          }
+
+          registration.pushManager.subscribe({
+            userVisibleOnly: true
```
romandev: We might need to take a server key?

client/push/push_manager.ts
```diff
@@ -19,25 +19,46 @@ declare var navigator: any;
...
 
-  constructor() {
+  constructor() {}
+
+  async isRegistered(): Promise<boolean> {
+    // Not implemented yet
+    return false;
+  }
+
+  async register(key: string): Promise<boolean> {
```
romandev: IMHO, returning endpoint is better.

client/push/push_manager.ts
```diff
@@ -19,25 +19,46 @@ declare var navigator: any;
...
+    // Not implemented yet
     if (navigator.serviceWorker) {
       navigator.serviceWorker.register('/push_service_worker.js')
         .then((registration: ServiceWorkerRegistration) => {
-          this.isRegistered_ = true;
+
+          if (!registration.pushManager) {
+            return false;
+          }
+
```
romandev: If have already subscription, it might be better to use it.

client/push/push_manager.ts
haeungun: Yes, we need it. 
Should I use the server key from the link below?https://github.com/haeungun/absolute/blob/06d7b71bdcfee70911d065d7ec024d5275ebe2fa/client/push/push_key.js

JayChl: If we use vapid, yes and it need converting to uint8array.
If we dont use vapid, it doesnt need. (Should use sender_id of manifest.json from firebase in case of gcm)
Recommand to use vapid

client/push/push_manager.ts
```diff
@@ -17,28 +17,51 @@
...
-  }
+  constructor() {}
 
   async isRegistered(): Promise<boolean> {
     // Not implemented yet
     return false;
   }
 
-  async register(key: string): Promise<boolean> {
+  async register(key: string): Promise<any> {
```
JayChl: What kind of key argument is going to be used for here? It seems there is no usage in this function.
FYI, Push subscription including endpoint will return by browser as you know, And subscription.keys also need by server so, should send to server

client/push/push_manager.ts
```diff
@@ -14,35 +14,64 @@
...
```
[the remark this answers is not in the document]
haeungun: The return type of an async function must be the Promise type.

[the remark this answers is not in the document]
romandev: @haeungun Ah sorry, so, shouldn't we return Promise<string> instead of Promise<any>.

client/push/push_manager.ts
```diff
@@ -17,28 +17,52 @@
...
+          })
+        })
+
+      })
+      .catch((error: Error) => {
+        // FIXME(zino): What should we do?
+      });
+    }
+
+    return 'register failed';
```
romandev: ```return null;``` is better I think.

client/push/push_manager.ts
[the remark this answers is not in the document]
romandev: OMG, I'm really sorry. The previous comments were broken.
Shouldn't we return ```Promise<string>``` instead of ```Promise<any>```.

PR #509, opened 2017-11-06
Implement skeleton codes for Cache APIs

sapzape:
This is an initial implementation to provide cache APIs to client app.
It consist of a simple structure. Need to implement it in detail later.

ISSUE=#534


sapzape:
I will check the cause of the CI build failure.

client/cache/cache_manager.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+declare var navigator: any;
+
+export default class CacheManager {
+    private isRegistered_: boolean = false;
+
+    constructor() {
+        if (navigator.serviceWorker) {
+            navigator.serviceWorker.register('/cache_service_worker.ts')
```
romandev: Should be cache_service_worker.js instead of cache_service_worker.ts.
The navigator.serviceWorker.register() is a function in browser side and the browser can not recognize TypeScript code.

sapzape:
I added some code that I did not need in the process of referring to other code. So I wrote it back to the base code only.

sapzape:
@romandev Ready to review.
I'll check the area where the error occurred in CI in the evening.

client/cache/cache_service_worker.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+self_.addEventListener('install', (event) => {
+    event.waitUntil(
```
romandev: Our indentation should be 2 instead of 4.

sapzape: I changed the indent from 4 to 2. I do not know why, what is the reason for 2 indentation?

sapzape:
PTAL :)

PR #510, opened 2017-11-08
Implement unregister for Push API

haeungun:
Implemented unregister function for push api.

ISSUE=none

### DIFF
--- a/client/push/push_manager.ts
+++ b/client/push/push_manager.ts
@@ -26,8 +26,6 @@ export default class PushManager {
   }
 
   async register(key: string): Promise<string> {
-    // Not implemented yet
-
     if (navigator.serviceWorker) {
       navigator.serviceWorker.register('/push_service_worker.js')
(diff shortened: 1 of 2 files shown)


client/push/push_manager.ts
```diff
@@ -67,6 +67,34 @@ export default class PushManager {
 
   async unregister(): Promise<boolean> {
     // Not implemented yet
+    if (navigator.serviceWorker) {
+      navigator.serviceWorker.register('/push_service_worker.js')
```
romandev: This is unregister() function but why are u registering service worker here again?

client/push/push_manager.ts
```diff
@@ -67,6 +67,34 @@ export default class PushManager {
 
   async unregister(): Promise<boolean> {
```
romandev: Can we remove this comment?

haeungun:
@romandev PTAL.
I did not separate indexeddb, but pass the test.

PR #513, opened 2017-11-08

llighter:
ISSUE= #503

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,6 @@
     "express": "^4.16.1",
     "gulp": "^3.9.1",
     "gulp-nodemon": "^2.2.1",
-    "gulp-shell": "^0.6.3",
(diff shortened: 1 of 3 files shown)


llighter:
I use 'child_process' module for running command shell. And I think 'child_process' is basic module in Node and also I could be used various way to run command shell.
My reference is [this](https://nodejs.org/api/child_process.html).
I'm sorry to be late on schedule.

package.json
```diff
@@ -9,12 +9,12 @@
     "@types/request": "^2.0.3",
     "@types/run-sequence": "0.0.29",
     "@types/supertest": "^2.0.3",
+    "child_process": "^1.0.2",
```
romandev: So, you don't need to add this, right? (if you are using built-in child_process module in Node.js)

llighter:

Issue #458, opened 2017-09-30
Implement base application framework.


romandev:
We should implement base application framework similar to JS version base[1] in v1.

[1] https://github.com/lunchclass/absolute/blob/absolute_v1/server/base/application.js

PR #515, opened 2017-11-09
Add bodyParser

hyungheo:
Add bodyParser Module and Test
It is needed for handling a request of post

ISSUE=#458

### DIFF
--- a/server/base/application.ts
+++ b/server/base/application.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import * as bodyParser from 'body-parser';
 import * as express from 'express';
 import * as path from 'path';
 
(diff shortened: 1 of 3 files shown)


server/base/application.ts
```diff
@@ -14,21 +14,32 @@
...
 import * as express from 'express';
 import * as path from 'path';
 
 export class Application {
   private static app: express.Application = express();
 
   public static async START(): Promise<void> {
-    await import('../example/example.router');
     this.app.use(express.static(path.join(__dirname, '../client')));
+    this.app.use(bodyParser.json({limit: '10mb'}));
```
romandev: Do we really need this limit?

server/base/application.ts
```diff
@@ -14,21 +14,32 @@
...
+      extended: true,
+    }));
+    await import('../example/example.router');
     this.app.listen(8090);
   }
 
   public static async START_FOR_TESTING(): Promise<express.Application> {
-    await import('../example/example.router');
     this.app.use(express.static(path.join(__dirname, '../../out/client')));
+    this.app.use(bodyParser.json({limit: '10mb'}));
```
romandev: ditto

server/example/example.router.ts
```diff
@@ -22,4 +22,15 @@ export class Test {
   public get(request: express.Request, response: express.Response): void {
     response.send('hello world');
   }
+  public post(request: express.Request, response: express.Response): void {
+    if (request.body) {
+      if (request.body.userId === 'absolute') {
```
romandev: Why userId? I think we need a better name for example.

hyungheo:
Express (>4.0)  needs body-Parser to parse request.body which has a specific data such as  json or x-www-form-urlencoded when it's handling  a POST.
I'll include some POST-API

romandev:
@hyungheo I understand your words. Thank you for explanation.
So, please update your commit description and title in more details (applying your previous comment).

server/base/application.ts
hyungheo: JSON is limited under 100KB as default.
Default size is not enough to handling big-data through JSON.

romandev: So, is there any use case in this patch?
If you have an use case, then please add it.
Otherwise, IMHO, remove it.

PR #516, opened 2017-11-09

llighter:
ISSUE= #503

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -14,18 +14,18 @@
  * limitations under the License.
  */
 
+import * as childProcess from 'child_process';
 import * as del from 'del';
 import * as gulp from 'gulp';
 import * as nodemon from 'gulp-nodemon';
 import * as runSequence from 'run-sequence';
 import tslint from 'gulp-tslint';
-import * as shell from 'gulp-shell';
 import * as tsc from 'gulp-typescript';
 
 const webpack = require('webpack-stream');
 
 gulp.task('default', (callback) => {
-  runSequence('webpack', 'build_server', 'run_server', callback);
+  runSequence('webpack', 'build_server', 'run_server', 'start-mongo', callback);
 });
 
 gulp.task('lint', () => {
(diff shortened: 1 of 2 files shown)


gulpfile.ts
```diff
@@ -19,13 +19,13 @@ import * as gulp from 'gulp';
 import * as nodemon from 'gulp-nodemon';
 import * as runSequence from 'run-sequence';
 import tslint from 'gulp-tslint';
-import * as shell from 'gulp-shell';
 import * as tsc from 'gulp-typescript';
 
 const webpack = require('webpack-stream');
+const { exec } = require('child_process');
```
romandev: Can we use `import * as childProcess from 'child_process';`

gulpfile.ts
```diff
@@ -49,7 +49,7 @@ gulp.task('lint:fix', () => {
     }));
 });
 
-gulp.task('test', ['webpack'], shell.task('jest'));
+gulp.task('test', ['webpack'], runCommand(`jest`));
```
romandev: Can we just use ' instead of `?

.gitignore
```diff
@@ -3,4 +3,5 @@ node_modules/
 out/
 third_party/mongodb/
 third_party/node/
-.DS_Store
\ No newline at end of file
+.DS_Store
+package-lock.json
```
romandev: Also, please sort this in alphabetical order.
BTW, do we really need this? I think this file will be created because you added child_process module to package.json.
If you don't add the module to package.json, I guess this might be unnecessary.

gulpfile.ts
llighter: OK, I'll change this.

gulpfile.ts
llighter: Copy that!

.gitignore
llighter: You're right.
It's because I'm using over 5 version of npm.
I'll get rid of it.

llighter:

PR #517, opened 2017-11-10
Add new gulp task to 'gulpfile.js' for running mongo daemon

llighter:
Applied the mongo daemon startup option to 'gulpfile.ts'.

ISSUE=#503

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -14,18 +14,18 @@
  * limitations under the License.
  */
 
+import * as childProcess from 'child_process';
 import * as del from 'del';
 import * as gulp from 'gulp';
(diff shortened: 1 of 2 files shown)


romandev:
@llighter Thank you for hard work.
This patch looks good to me. I'll merge this after passing all tests.

BTW, I'll point minor things out.
- According to the contribution guide, we don't have to use periods(.) in the title.
  - https://github.com/lunchclass/absolute/blob/master/docs/contributing.md#commit
- If possible, we should use imperative sentense.
  - For example, you write "I applied something blah blah.." but we had to write "Applied something blah blah".

Thank you very much!

PR #523, opened 2017-11-12
Fix a build error

romandev:
Sometime build error occurs as follows:
An async function or method must return a 'Promise'. Make sure you have
a declaration for 'Promise' or include 'ES2015' in your `--lib` option.

ISSUE=none


nadongguri:
@romandev, is this issue still happened?

PR #524, opened 2017-11-12
Add application manager for client

elliekwon:
Application manager for client created. This pull request has only structure and test code for checking chrome browser.

### DIFF
--- a/client/application/application_manager.ts
+++ b/client/application/application_manager.ts
@@ -14,14 +14,12 @@
  * limitations under the License.
  */
 
-import PushManager from './push/push_manager';
-import Notification from './notification/notification_manager';
-import AppManager from './application/application_manager';
-import IndexedDB from './indexeddb/indexeddb';
+declare var navigator: any;
 
-export default class absolute {
-  static push: PushManager = new PushManager();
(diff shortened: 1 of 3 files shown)


client/application/application_manager.ts
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export default class ApplicationManager {
+
+  async isCromeBrowser(): Promise<boolean> {
+    var agent = navigator.userAgent.toLowerCase();
+    if(agent.indexOf('chrome') != -1){
```
romandev: This is not perfect. For example, Samsung Internet, Opera, Whale browser will also include the `chrome` keyword in UA string.
Also, Android system WebView also includes `chrome` keyword in UA string.
You will have to investigate more things.

nadongguri: @elliekwon, if you are in trouble with this, take a look at the previous commit. [link](https://github.com/lunchclass/absolute/blob/hackathon/client/deeplink/deeplink.js)

PR #529, opened 2017-11-13
Add @nadongguri and @yjaeseok to CODEOWNERS

romandev:
@nadongguri has been contributing Absolute project for a long time.
@yjaeseok doesn't have a lot of landed patches, but since v2, he has
actively been reviewing most of PRs.

We believe that they are understand our process well.

ISSUE=none


yjaeseok:
Thanks! @romandev

nadongguri:
I will try hard because it will be a burden because the role given is important.

PR #530, opened 2017-11-16
Fixed wrong url

haeungun:
`application.ts` has a wrong url path.
I fixed it to be the correct path.

ISSUE=none

### DIFF
--- a/server/base/application.ts
+++ b/server/base/application.ts
@@ -25,7 +25,7 @@ export class Application {
   private static app: express.Application = express();
 
   public static async START(): Promise<void> {
-    this.app.use(express.static(path.join(__dirname, '../client')));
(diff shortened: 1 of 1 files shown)


server/base/application.ts
```diff
@@ -25,7 +25,7 @@ export class Application {
   private static app: express.Application = express();
 
   public static async START(): Promise<void> {
-    this.app.use(express.static(path.join(__dirname, '../client')));
+    this.app.use(express.static(path.join(__dirname, '../../client')));
```
romandev: Shouldn't be `../../out/client`?

haeungun: `application.ts` will be build into `out/server/application.js`
So static path should be set by considering runtime path.

haeungun:
Okay, I'll update it.

Issue #533, opened 2017-11-17
Add support for lint in client/


romandev:
Current lint command only affects server/ directory.
But we have already much codes in client/ directory as well.
So, we should make the lint command apply to client/ directory as well and fix all lint error.

sapzape:
I'll take this issue :)

romandev:
@sapzape Go ahead! Thanks :)

PR #535, opened 2017-11-17
Modified lint command to accept multiple arguments

sapzape:
This change fixes the lint command apply to client/ directory as well.
ISSUE=#533


sapzape:
It was wrong. I will finish processing and pull request again.

Issue #500, opened 2017-10-28
Designing REST API for the client side in push-notification


GarlicB:
Maybe i did not understand much of the requirements. Please give quick feedback.
Anyway, i referenced [kakao developers](https://developers.kakao.com/docs/restapi/push-notification)

# endpoint

**Host**: .../push (just example)
**Data format** : application/json

**GET   /push**
This page lets you know if the current user has a token for the push notification.

**POST /push/register**
**POST /push/unregister**
In */push* page, user can register or unregister with the same button. At this time, the necessary data is added to the body and processed as POST.

**POST /push/event**
This is the page that sends a notification to the user.

romandev:
@GarlicB Thank you for this.
But you missed some parameter information. Please add the info in more details.
Also, I'm not sure if we need /push/event API in server side.

/cc @daehyunjimmy @JayChl

GarlicB:
@romandev OK i'll add more info.

PR #539, opened 2017-11-19
Implement skeleton codes in server/push

GarlicB:
To avoid conflict, updated the skeleton codes again at the most recent
commit phase.

ISSUE=#500


GarlicB:
oops, soon i’ll fix lint err

nadongguri:
Minor one, would you change the position of the sentences? The description should be first then ISSUE line follows. :)

GarlicB:
@nadongguri, OK. i changed the position

PR #540, opened 2017-11-24
Fixed some lint errors in client directory using `lint:fix`

sapzape:
Performed the `lint:fix` command. That fixes some lint errors in
`client/` directory.

ISSUE=#533


romandev:
lgtm;

I modified your commit description as follows:
  - DO NOT USE `I did something blah blah`. If possible, it's better to use imperative mood.
  - Each line should be < 72 byte.
  - Need a blank line your commit description and `ISSUE=#533`

PR #543, opened 2017-12-02
Handle errors about third party download and extract

cs-lee:
If script fail to download node binary, it should stop
and print right error log. This patch do it.
Additionally this patch add one function to print red color error log.

ISSUE=none

### DIFF
--- a/bootstrap/common/sync_third_party.sh
+++ b/bootstrap/common/sync_third_party.sh
@@ -29,16 +29,7 @@ function sync_node() {
     darwin_x86_64) local target_url="/node-v6.9.5-darwin-x64.tar.gz" ;;
   esac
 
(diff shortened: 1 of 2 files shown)


nadongguri:
@cs-lee, LGTM. Would you share how to fallback to error 2 or 3 cases?

cs-lee:
case 2 :
- 1. Delete third_party/node , third_party/mongodb (For download again binary)
- 2. Disconnect lane cable and do ./absolute test 

case 3 : 
- If sync_third_party.sh - sync_node this function send invalid file url to sync_third_party function.
tar or unzip command will fail.

cs-lee:
@romandev I applied your review comments. Please review.

